PR #1, opened 2017-12-07
[#153072577] First steps to admin app: Base rails app with OAuth

chrisfarms:
# What

We're building a web UI for the paas.

This is the first steps to that end:

* A base Rails app to build upon
* Login/Authentication via UAA/OAuth
* A basic CF API Client
* A page that lists org names as an example
* Some tests

# How to review

* Code review
* Ensure the tests make sense and pass (see [README](README.md))
* Run the app and test logging in as a user:
    * You will need to first deploy: https://github.com/alphagov/paas-cf/pull/1135
    * You will then we able to run the app locally (see [README](README.md)) or push to your dev (if you set all the required environment variables manually).

# Who can review

Not @chrisfarms nor @henrytk nor @camelpunch nor @carolinegreen


dcarley:
I've enabled Travis for this repo and triggered a build through their admin UI.

chrisfarms:
I have added a .travis.yml that should trigger the (non-integration) tests

chrisfarms:
Ignore that. Travis seems to have fixed itself.

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <head>
+    <title>PaasAdmin</title>
+    <%= csrf_meta_tags %>
+
+    <%= stylesheet_link_tag    'application', media: 'all' %>
+    <%= javascript_include_tag 'application' %>
+  </head>
+
+  <body>
+    <img src="https://static1.squarespace.com/static/59698a7d29687fd47a2a7c52/599a2c9837c581a0f21a7030/5998fbff17bffc5c4b7d0f04/1503700549384/construction_worker.jpg" border="0" width="125">
```
dcarley: Should we vendor this rather than hotlink?

.envrc
```diff
@@ -1,6 +1,8 @@
...
-export OAUTH_CLIENT_SECRET="oohilovesecrets"
-export AUTH_SERVER_URL="https://login.bosh-lite.com"
-export TOKEN_SERVER_URL="https://uaa.bosh-lite.com"
-export SKIP_TLS_VERIFICATION=true
+export PATH="./bin:$PATH"
+export CF_API_ENDPOINT="https://api.${DEPLOY_ENV}.dev.cloudpipeline.digital"
+export CF_AUTH_ENDPOINT=$(cf curl /v2/info | jq -r .authorization_endpoint)
+export CF_TOKEN_ENDPOINT=$(cf curl /v2/info | jq -r .token_endpoint)
+export CF_CLIENT_ID="paas-admin"
+export CF_CLIENT_SECRET=$(aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/cf-secrets.yml - | awk '/uaa_clients_paas_admin_secret/ { print $2 }')
```
dcarley: Should we include things that are so specific to our development environments? Do we want other people to use the app in the future?

bin/bundle
```diff
@@ -1,3 +1,105 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
```
dcarley: Do you know why we need a binstub for Bundler?

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    if not valid_token?
+      session[:access_token] = nil
+      redirect_to('/auth/cloudfoundry')
+    end
+  end
+
+  def valid_token?
+    token = session[:access_token]
+    return false if token.nil?
+    return false if token.empty?
```
alext: These 2 lines could be replaced with a `token.blank?` test.

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    decoded_token = CF::UAA::TokenCoder.decode(token, :verify => false)
+    return false if not decoded_token.has_key? "exp"
```
alext: It would be better to avoid the double-negative, and use `unless` here.

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    if decoded_token["exp"] > Time.now().to_i
+      return true
+    end
+    return false
```
alext: These 4 lines could be simplified to:
```ruby
return decoded_token["exp"] > Time.now().to_i
```

app/views/layouts/application.html.erb
alext: :+1: We should be hosting that ourselves - makes us better Internet citizens, and also we can't be sure of the longevity of that link etc.

dcarley:
GitHub has collapsed some of my comments possibly because of the commit that I commented them from, but I think they're all still relevant.

Oh, one last thought. The story summary in Pivotal lists this as one of the follow-ups:

> Rubocop checking/linting via travis

I reckon that we should do this now because it doesn't have an obvious home in another story and it'll be easier to do the earlier we do it.

config/initializers/cf_client.rb
```diff
@@ -0,0 +1,12 @@
+require 'cf'
+
+Rails.configuration.api_endpoint = ENV.fetch("CF_API_ENDPOINT")
```
alext: I think this `Rails.configuration` entry would be better named `cf_api_endpoint` to avoid any possible ambiguity.

spec/spec_helper.rb
```diff
@@ -0,0 +1,98 @@
...
+  # This option will default to `:apply_to_host_groups` in RSpec 4 (and will
+  # have no way to turn it off -- the option exists only for backwards
+  # compatibility in RSpec 3). It causes shared context metadata to be
+  # inherited by the metadata hash of host groups and examples, rather than
+  # triggering implicit auto-inclusion in groups with matching metadata.
+  config.shared_context_metadata_behavior = :apply_to_host_groups
+
+# The settings below are suggested to provide a good initial experience
+# with RSpec, but feel free to customize to your heart's content.
+=begin
```
alext: I think it's worth enabling all these default settings (ie remove the `=begin` and `=end` lines). I think `disable_monkey_patching!` is a setting it's particularly good to have in place.

chrisfarms:
RE: "scenarios where [auth] it might be bypassed or broken"

The `pass-admin` UAA client has no authorities of it's own, and is currently only allowed to request authorization for set of read-only scopes from a user. The app cannot _do_ anything on it's own even if authentication was bypassed, since the token is just passed directly to `ccapi` to make decisions about what data the token is authorized to view.

bin/bundle
chrisfarms: something something bug in bundler todo with binstubs .... /cc @camelpunch

camelpunch: I don't know why bundler (or Rails) introduced a binstub for bundler itself. I assume it's to lock the version of bundler for the app.

The bug that @chrisfarms is referring to is here, but not sure if this is related to the question: https://github.com/bundler/bundler/issues/6149

spec/features/list_orgs_spec.rb
```diff
@@ -8,20 +8,22 @@ def app
...
-      expect(last_response.header['Location']).
-        to eq(oauth_redirect_url)
+      actual = last_response.header['Location']
+
+      expected = [
+        ENV.fetch('AUTH_SERVER_URL'),
+        '/oauth/authorize',
+        "?client_id=#{ENV.fetch('OAUTH_CLIENT_ID')}",
+        '&response_type=code',
+        "&redirect_uri=#{CGI.escape("http://example.org/auth/cloudfoundry/callback")}",
```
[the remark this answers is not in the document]
chrisfarms: We discarded this technique for testing the auth once we realized all we were testing was OmniAuth and that it was near imposible to test without mocking UAA too .... we then found OmniAuth has some documented ways of mocking the auth and used that method instead.

Essentially we trust that OmniAuth works and currently have no integration tests to prove that (once we have a CI build pipeline we would probably want to add those sorts of integration/acceptance tests).

app/views/layouts/application.html.erb
chrisfarms: I love your optimism that Construction Duck will last persist long enough to be worried about this.

spec/features/list_orgs_spec.rb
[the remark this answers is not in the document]
camelpunch: To answer @dcarley's question: it's the [default 'fake' app host](https://github.com/rack-test/rack-test/blob/43a3ccffffe541abc6bf18c11e5be920108b4b38/lib/rack/test.rb#L13) that rack-test uses.

.envrc
chrisfarms: I have mixed feelings.

Setting up these variables for running in dev is painful. But I understand this is so specific to us that would be annoying for outsiders.

Henry created some scripts for running (to replace/augment .envrc) ... maybe we should document the "standard" way ie `CF_AUTH_ENDPOINT=xx CF_TOKEN_ENDPOINT=yy rails s` in the README for all to enjoy ... but then have some additional documentation for paas-specific-ness?

Gemfile
```diff
@@ -10,7 +10,7 @@ end
 gem 'rails', '~> 5.1.4'
 gem 'puma', '~> 3.7'
 gem 'uglifier', '>= 1.3.0'
-gem 'omniauth-uaa-oauth2'
+gem 'omniauth-uaa-oauth2', github: 'cloudfoundry/omniauth-uaa-oauth2'
```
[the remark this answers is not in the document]
camelpunch: I think this is concerning. We should complain, or do a spike into using vanilla oauth with https://github.com/omniauth/omniauth-oauth2

bin/bundle
dcarley: I'm inclined to delete it rather than worry about when we can update/remove the backported fix. It doesn't seem as important the `rails` binstub, which appears to be a standard for Rails apps.

spec/features/list_orgs_spec.rb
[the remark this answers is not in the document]
dcarley: Thanks and thanks 👍

.envrc
dcarley: That sounds good. I definitely appreciated having an easy way to run it against my dev environment, but I think it would be easier to understand if we provided generic instructions foremost.

bin/bundle
camelpunch: The fix was primarily for stubs like `rspec`, which are used a lot. I think the only risk here is if someone:

a) Decides to run e.g. `bundle binstubs rspec-core` and
b) Commits the change

Otherwise if we're concerned about bundler versions, we should downgrade to 1.15, as [one of the comments on the Rails bug thread suggested](https://github.com/rails/rails/issues/31193#issuecomment-346116270), because 1.15 doesn't generate incompatible binstubs by default (1.16 is fine [if you also supply `--force`](https://github.com/rails/rails/issues/31193#issuecomment-347784561))

Gemfile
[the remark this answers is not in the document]
dcarley: I hadn't notice this before, which is more concerning: https://github.com/cloudfoundry/omniauth-uaa-oauth2/issues/10

@alext any thoughts? Should we merge this as-is and have a story to look at changing it afterwards?

app/controllers/orgs_controller.rb
```diff
@@ -1,5 +1,5 @@
 class OrgsController < ApplicationController
   def index
-    redirect_to Rails.configuration.x.oauth.redirect_url
```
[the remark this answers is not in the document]
camelpunch: (this change has been rebased away)

Gemfile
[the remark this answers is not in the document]
dcarley: @camelpunch @alext and I talked about this: we decided to leave it as-is for this PR and @camelpunch will write a follow-up story to investigate `omniauth/omniauth-oauth2`.

Gemfile
```diff
@@ -7,24 +7,25 @@ git_source(:github) do |repo_name|
...
   gem 'capybara'
+  gem 'pry-byebug'
   gem 'rack_session_access'
+  gem 'rspec-rails'
 end
 
 group :development do
-  gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.2'
+  gem 'rubocop', '~> 0.51.0', require: false
```
dcarley: We should use `govuk-lint` as a gem and `govuk-lint-ruby` as the command, because then we don't need to vendor all of it's `.rubocop.yml` and it will get updated when we pull in new versions of the gem.

You can refer to `alphagov/paas-cf` for an example:

- https://github.com/alphagov/paas-cf/blob/c11494efc27c03ad61a4a2ced9dce548315b5da2/Gemfile#L5
- https://github.com/alphagov/paas-cf/blob/c11494efc27c03ad61a4a2ced9dce548315b5da2/Gemfile#L5
- https://github.com/alphagov/paas-cf/blob/c11494efc27c03ad61a4a2ced9dce548315b5da2/.rubocop.yml

chrisfarms: I'm not sure if that's the best move tbh. I can see two issues with that:

1) Anyone who has an editor/plugin that does rubocop linting for them would not understand the "govuk-lint-ruby" command and would expect to find the rubocop.yml file in the root. I find it very useful to get the cop feedback as I'm typing and it would be a real shame to lose that.

2) The govuk-lint version is not a perfect fit... I had to enable the rails specific cops and had to modify the include/excludes to ensure that certain cops are excluded from some rails specific files (large blocks in environment configs are expected for one) and needed to exclude the vendor dir so travis didnt attempt to lint the entire vendor bundle.

chrisfarms: ... just talking with @alext  ... we think we can have our cake and eat it ... *investigating*

dcarley: The README says that there's a `-R` flag to enable Rails cops:

- https://github.com/alphagov/govuk-lint#ruby

I think it's worth trying to find solutions to because people have already done the hard work of maintaining an acceptable config ❤️

chrisfarms: 🙌  ...sorted!

Found a way to inherit the config from the govuk-lint gem so that standard `rubocop` commands work as expected and we get the best of both worlds.

dcarley:
Nice!! Thanks 👍

PR #3, opened 2018-02-12
Blackbox testing for server/logging 

chrisfarms:
## What

We should be testing that server process itself (`main.js`) actually does what we expect. We want to ensure:

* It starts
* It outputs structured logs to stdout
* It can be configured via the environment
* It gracefully shutsdown on `SIGINT` and `SIGTERM`

This change factors out the main server start/stop/update handling into it's own module so it can be tested and makes main.js responsible for setting up the configuration of the app (the logger, port etc).

To make the experience of working with the tests better the `--jobs` flag is used to run tests in parallel, and `npm run watch` task is added to take advantage of incremental builds.

## How to review

* Code review
* Ensure tests are ok and server starts

## Who can review

Not @chrisfarms


package.json
```diff
@@ -7,8 +7,9 @@
   "repository": "https://github.com/alphagov/paas-admin",
   "scripts": {
     "build": "npm run -s clean && webpack --config config/webpack.config.js",
-    "test": "(npm run -s test:unit -- --cov && npm run -s lint) && echo '\n✓ ALL TESTS PASSED!'",
-    "test:unit": "ENABLE_TESTS=true npm run -s build && tap --100 --no-coverage --nyc-arg='--instrument=false' --nyc-arg='--all' dist/*.test.js",
+    "watch": "ENABLE_WATCH=true ENABLE_TESTS=true npm run -s build",
```
paroxp: Shouldn't this command also do `webpack --watch`?

package.json
```diff
@@ -17,8 +18,8 @@
     "fix:js": "xo --fix ./src",
     "fix:sass": "sass-lint --fix ./src",
     "fix:vulnerabilities": "snyk protect",
-    "start": "if [ \"$NODE_ENV\" = 'production' ]; then npm run start:production; else npm run start:dev; fi",
-    "start:dev": "NODE_ENV=development ENABLE_WATCH=true ENABLE_SERVER=true npm run build | pino",
+    "start": "export PORT=${PORT-3000}; if [ \"$NODE_ENV\" = 'production' ]; then npm run -s start:production; else npm run -s start:dev; fi",
```
paroxp: Shouldn't it be `${PORT:-3000}`? 🤔 

It's getting a little complicated. I wonder when are we going to say "let's use Makefile" :D

package.json
```diff
@@ -57,6 +58,7 @@
     "webpack": "4.0.0-beta.0",
     "webpack-cli": "2.0.4",
     "webpack-node-externals": "1.6.0",
+    "webpack-shell-plugin": "^0.5.0",
```
paroxp: Should we pin it? :D

config/tests.config.js
```diff
@@ -1,27 +1,45 @@
...
+    args = `--100`;
+  }
+
+  // Add plugin to execute test runner
+  cfg.plugins.push(new WebpackShellPlugin({
+    onBuildExit: [{
+      command: `sh`,
+      args: [
+        `-c`,
+        `./node_modules/.bin/tap ${args} --nyc-arg='--instrument=false' --reporter classic -J ./dist/*.test.js && echo '\n✓ ALL TESTS PASSED!' ${onFail}`
```
paroxp: Why are we forcing a success?

package.json
chrisfarms: we are not using the `--watch` arg ... we are setting `watch: true` in the webpack config to because we need to do more than just enable watch mode, we need to alter how tests are run (should not explode and bail out on failure when watching) and alter how the server runs (should hot reload modules when watching). We're not using _any_ of the webpack cli args it's all done through the config.

package.json
chrisfarms: Both are valid in this case ... I think... `${PORT-3000}` will use 3000 if PORT is 'unset' ... which allows setting it to nothing `PORT=` to get the random port behaviour if you were so inclined

config/tests.config.js
chrisfarms: when running in watch mode ... you want the tests to run after each time the source changes ... if this script returns with a failure status then it will cause the webpack watch process to exit, which means you'd have to restart the process every time you cause a test failure which kinda defeats the point.

When running _without_ the watch loop we _do_ want the tests to exit with a nasty status to ensure we get test failures.

package.json
paroxp: TIL.

config/tests.config.js
paroxp: Sounds good! :tada:

PR #7, opened 2018-02-22
[#154558689] Implement oauth2

paroxp:
## What

We're implementing a OAuth2 solution in pazmin. We're using `passport` library to handle most of the magic for us. 

We've covered all code used, but there are things we're not testing due to the approach taken in the beginning of this story.

We'd like to test whether a cookie that has been tampered with, does not match with the signature meaning the app would logout the user.

Rafal really doesn't like "magical" middleware that simply applies logic to any route that follows it's declaration. Bother him about details. - This is still on the table and may end up being improved in a follow up story. UNLESS the reviewer agrees with Rafal.

We've also encountered a problem that a thrown error will not stop the tests but instead will hang them. We agreed this will be fixed in a follow up story.

Along there are some typo corrections and cleaning up/fixing some other bugs introduced in previous stories.

## How to review

- Code review
- Run the tests `npm test` (will be run by travis)
- Run the application 
  ```sh
  OAUTH_AUTHORIZATION_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/authorize" OAUTH_TOKEN_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/token" OAUTH_CLIENT_ID="rafal-login" OAUTH_CLIENT_SECRET="secret" SERVER_ROOT_URL="http://localhost:3000" npm start
  ```
- Attempt to preview `/orgs` - you should be redirected to Rafal's UAA
- Ask nicely for the password or an account
- Log in - you should be redirected back to the `/orgs` and see the actual page
- Check if the token is valid and we can pull orgs from CF API - we've prepared a TMP commit 
1f2382e 
- Confirm app displays some orgs user has access to `/tmp-orgs`
- Remove the TMP commit once not needed.

🐝 🐝 🐝 🐝  You should remove the TMP commit. Also, the travis fails purely because the code is not covered in tests.

## Who can review

Neither @keymon, @46bit, @carolinegreen nor myself.


src/app/app.js
```diff
@@ -1,32 +1,58 @@
...
 
-  if (logger) {
-    app.use(pinoMiddleware({logger}));
-  }
+  app.use(pinoMiddleware(config.logger));
+
+  app.use(cookieSession({
+    name: 'pazmin-session',
+    keys: [config.sessionSecret],
+    secure: config.isProduction,
```
chrisfarms: It's useful to be able to run the server locally with NODE_ENV=production to test minor build differences (like with minification/performance enhancements/whatever)... ie `npm run start:production`

by basing the `secure` on the NODE_ENV value I think this makes it currently impossible to run the "production" build locally.

Perhaps an `INSECURE=true` option to disable https-only cookies (and maybe other things like it) could work instead, so we could modify the `npm start` helper to set the `INSECURE=true` flag and maintain the ability to run the production build.

src/app/app.js
```diff
@@ -1,32 +1,58 @@
...
 
-  app.use('/orgs', orgs);
   app.use('/', home);
 
+  // Authenticated endpoints follow
+  app.use(auth(config));
+
+  app.use('/orgs', orgs);
+
+  app.get('/tmp-orgs', (req, res) => {
```
chrisfarms: Should we be removing this before merging? ... I know it's temporary junk but could we not replace the hard coded url with a `${CF_API_ENDPOINT}/v2/organisations` or something and leave it as an example?

src/app/app.js
```diff
@@ -1,32 +1,58 @@
 import express from 'express';
+import cookieSession from 'cookie-session';
 import helmet from 'helmet';
 import pinoMiddleware from 'express-pino-logger';
 import compression from 'compression';
 import staticGzip from 'express-static-gzip';
+import fetch from 'node-fetch';
```
chrisfarms: `node-fetch` won't play nicely with webpack when it does a production build 😭  

... it looks like this is only used in the "TMP" commit, so it's probably not something to worry about right now 

... IIRC [axios](https://github.com/axios/axios) is quite a popular promisey/http lib

src/app/app.js
paroxp: Initially I had it in orgs an was keen to keep as an example for future stories. I believe @keymon had some opinions as to why we shouldn't do that.

src/app/app.js
paroxp: Yeah, it's part of our TMP commit just to prove a point. It should go and ideally be replaced by node-cfapi of some sort.

src/main.test.js
```diff
@@ -51,7 +77,7 @@ t.test('should exit with non-zero status on error (invalid PORT)', t => {
 
 async function run(env = {}) {
   return new Promise((resolve, reject) => {
-    const proc = spawn('node', ['./dist/main.js'], {env, shell: '/bin/bash'});
+    const proc = spawn(process.argv0, ['./dist/main.js'], {env});
```
chrisfarms: nice 👍

package.json
```diff
@@ -78,6 +85,12 @@
           "unicorn/no-process-exit": "off"
         }
       },
+      {
+        "files": "src/**/*.test.js",
+        "rules": {
+          "camelcase": "off"
```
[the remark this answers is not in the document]
paroxp: I would agree the comment approach is better, however it looks like we'd be doing it a lot in tests.

I'm unsure about this in all honesty. Maybe instead of camelcase, we should disable the need of removing quotes? for instance `{'this_is_prop':'test'}`?

src/app/app.js
chrisfarms: It's not massively important ... but in one glace at that function I worked out:

"oh so the auth stuff must give me a nice `req.session.passport.user` thing that probably contains the token"

if it goes away we should probably at least say something in the README like "our authentication middleware adds the `req.session.passport.user` property which contains an access_token

src/app/app.js
paroxp: Personally, I dislike the fact we need to pass bunch of env variables on start xD

```sh
OAUTH_AUTHORIZATION_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/authorize" OAUTH_TOKEN_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/token" OAUTH_CLIENT_ID="rafal-login" OAUTH_CLIENT_SECRET="secret" SERVER_ROOT_URL="http://localhost:3000" npm start
```

I'm use to `.env` files, but I don't want to introduce that here as I know people would disagree with me.

I agree with your point though, will do.

chrisfarms: Given the nature of this app we can't really avoid it ... however technically we can get the AUTH_ENDPOINT and TOKEN_ENDPOINT from the CF_API_ENDPOINT via /v2/info ... so that might make things a bit tidyier

package.json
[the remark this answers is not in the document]
chrisfarms: Cool ... leave it as-is 👍 ... I'll just have to restrain myself from adding overrides 😈

src/app/app.js
```diff
@@ -1,32 +1,58 @@
...
+  }));
 
   app.use('/assets', staticGzip('dist/assets', {immutable: true}));
   app.use(compression());
 
   app.use(helmet());
   app.use(helmet.contentSecurityPolicy(csp));
 
-  app.use('/orgs', orgs);
   app.use('/', home);
```
chrisfarms: I think the `home` route should be behind the authentication ... we have no need for any "unauthenticated" routes yet

src/auth/auth.js
```diff
@@ -0,0 +1,66 @@
...
+  passport.serializeUser((user, cb) => {
+    cb(null, user.accessToken);
+  });
+
+  passport.deserializeUser((accessToken, cb) => {
+    cb(null, {accessToken});
+  });
+
+  app.get('/auth/login', passport.authenticate('oauth2', {
+    scope: [
```
chrisfarms: You do not need to pass the scopes here. If you simplify it to:

```
app.get('/auth/login', passport.authenticate('oauth2'));
```

...then it will pull the scopes from the uaa-client's list of allowed scopes, which will mean you don't have to duplicate the scopes here as well as in the uaa-client config.

src/auth/auth.js
```diff
@@ -0,0 +1,66 @@
...
+    }, `Authentication successful, redirecting back to ${redirection}`);
+  });
+
+  app.get('/auth/logout', (req, res) => {
+    req.logout();
+    res.redirect('/');
+  });
+
+  app.use((req, res, next) => {
+    if (req.isAuthenticated()) {
```
chrisfarms: low priority: 

maybe we could set `req.accessToken = req.session.passport.user` here as a shortcut to the rather verbose `req.session.passport.user` for consumers of the auth middleware.

It's also a bit nicer to be able to do:

```
if (req.accessToken) { ... }
```

than 

```
if(req.session && req.session.passport && req.session.passport.user) { ... }
```

src/app/app.js
paroxp: This is really bad excuse, but: Moving this, will fail the tests. We should probably fix that by coming up with a better way of handling the token in tests. But I don't want to block this story on that. Although it introduces this problem, the idea of homepage is still undecided. Let's cross that bridge when we get to it?

PR #10, opened 2018-02-28
[#155508629] Refactor tests

paroxp:
## What

With the current use of `supertest` we can experience some unclear
issues which result in hanging the tests without any notification as to
what went wrong. For instance, the following code:

```js
request(app)
  .get('/')
  .expect(nonExsistingVariable);
```

Should in fact throw an exception and stop the tests. The reality, is
that the exceptions are being caught and could then be handled with
`.catch()` method.

We thought it's unreasonable for us to add that after every `.except()`,
therefore we decided to go with different approach.

It proved difficult to replace `supertest` with another solution, due to
the goodness it provides when setting up and tearing down the
application on the go. We decided to refactor it in a way to get the
best of both worlds. Keep the `supertest` magic and not use the bugged
`.except()` function.

We're forcing the request to finish, by prefixing it with the `await`
key word and then taking the response to be used with the default tap
functionality.

!!! This does not get rid of the issue... It should however discourage
people from using or knowing of the faulty functionality.

## How to review

- Code review - The code should be clear. You should know what's going on.
- Run the tests - `npm test` - also run by travis
- Advice perhaps if you think it's insufficient

## Who can review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4400,16 +4400,6 @@
         "homedir-polyfill": "1.0.1"
       }
     },
-    "expect-cookies": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/expect-cookies/-/expect-cookies-0.1.2.tgz",
-      "integrity": "sha1-Gsc3T+RtiEfcnm4YEUZ3B72U+f4=",
-      "dev": true,
-      "requires": {
-        "cookie-signature": "1.0.6",
-        "should": "7.1.1"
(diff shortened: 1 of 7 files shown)


paroxp:
FINE... Be like that.

keymon:
This message drove me to a interesting reading of [*'To not be' or 'not to be'*](http://www.learnersdictionary.com/qa/to-not-be-or-not-to-be)... 

I merge now :)

PR #12, opened 2018-03-02
[#154558816] Setup API communication

paroxp:
## What

We're going to create our own mini client, that will do all the necessary curl requests for us and return simple values.
The ones available for grabs, are out of date and could go distinct.

We're adding [`axios`](https://github.com/axios/axios) to our dependencies in order to make some requests across the web. We could be using the [`http`](https://nodejs.org/api/http.html) but we'd like to have an ability of promises.

We'd like to test it out by executing a call against the orgs endpoint to gather some data.

We're not doing a templating around it in this story, so these should be purely printing JSON in HTML.

## How to review

- Code review + sanity check
- Decide whether the client is sufficient in the current form
- See if you can obtain orgs under `/orgs` - use Rafal's setup and ask for admin password

## Who can review

Neither @camelpunch nor myself.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -53,7 +53,8 @@ const config = {
   oauthTokenURL: expectEnvVariable('OAUTH_TOKEN_URL'),
   oauthClientID: expectEnvVariable('OAUTH_CLIENT_ID'),
   oauthClientSecret: expectEnvVariable('OAUTH_CLIENT_SECRET'),
-  serverRootURL: process.env.SERVER_ROOT_URL || 'http://localhost:' + (process.env.PORT || '3000')
+  serverRootURL: process.env.SERVER_ROOT_URL || 'http://localhost:' + (process.env.PORT || '3000'),
+  cloudFoundryAPI: expectEnvVariable('API_URL')
 };
 
 main(config).then(onShutdown).catch(onError);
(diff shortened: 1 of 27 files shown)


src/main.js
```diff
@@ -53,7 +53,8 @@ const config = {
   oauthTokenURL: expectEnvVariable('OAUTH_TOKEN_URL'),
   oauthClientID: expectEnvVariable('OAUTH_CLIENT_ID'),
   oauthClientSecret: expectEnvVariable('OAUTH_CLIENT_SECRET'),
-  serverRootURL: process.env.SERVER_ROOT_URL || 'http://localhost:' + (process.env.PORT || '3000')
+  serverRootURL: process.env.SERVER_ROOT_URL || 'http://localhost:' + (process.env.PORT || '3000'),
+  cloudFoundryAPI: expectEnvVariable('API_URL')
```
henrytk: Before reading this diff I followed the README and got this error:

```
[2018-03-06T10:55:20.412Z] ERROR (57651 on gds3824.local): Expected environment variable "API_URL" to be set.
```

Could you add it to the README please?

paroxp: Good shout, will do!

henrytk:
Closing in favour of https://github.com/alphagov/paas-admin/pull/14 which contains all of these commits and one more.

PR #13, opened 2018-03-05
Make the Auth middleware check for expired token

paroxp:
## What

It turns out Passport has no idea we're dealing with JWT, meaning we
have to handle checking the expiration ourselves.

We're adding `jsonwebtoken` as a dependency for quick and easy decoding
of a token on a fly. It doesn't unfortunately allow us to validate the
token as it requires a secret to be fully verified.

Instead, we're decoding the Token and manually comparing the timestamp
with the current one to see if we've expired or good to go.

If the token has expired, we're going to kick the user back to the login
page for re-authentication.

## How to review

- Code sanity check
- Run tests (should be run by travis)
- Optionally log in to the app and wait (whatever it is) for the token to expire and see the action

### DIFF
--- a/src/auth/auth.js
+++ b/src/auth/auth.js
@@ -32,13 +32,7 @@ export default function authentication(config) {
   app.get('/auth/login', passport.authenticate('oauth2'));
 
   app.get('/auth/login/callback', passport.authenticate('oauth2'), (req, res) => {
-    const redirection = req.session.returnTo;
-    delete req.session.returnTo;
-
-    res.redirect(redirection || '/');
-    req.log.info({
-      returnTo: redirection
-    }, `Authentication successful, redirecting back to ${redirection}`);
+    res.redirect('/');
   });
 
   app.get('/auth/logout', (req, res) => {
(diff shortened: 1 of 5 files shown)


chrisfarms:
If we ever need to verify the token, there are [UAA endpoints that can do this](https://docs.cloudfoundry.org/api/uaa/version/4.8.0/index.html#check-token) but that would be horribly slow round trip ...so there is also an API to [get keys that can be used to verify token](https://docs.cloudfoundry.org/api/uaa/version/4.8.0/index.html#token-key-s) from UAA.

Checking the "exp" field and dropping the token without verification is definitely a good move 👍

src/auth/auth.js
```diff
@@ -46,7 +47,7 @@ export default function authentication(config) {
   });
 
   app.use((req, res, next) => {
-    if (req.isAuthenticated()) {
+    if (req.isAuthenticated() && !isExpired(req.session.passport.user)) {
```
chrisfarms: Maybe we should also also remove the cookie  whenever it is expired/invalid to avoid leaving dead "sessions" laying around in browser cookies?

ie something like... 

```
...
if (isExpired(token)) { 
res.headers[`Set-Cookie`] = 'token=deleted; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT';
res.redirect(req.url)
}
...
```

chrisfarms: Could we also set the cookie expiry date to the same as what we have in the jwt token so the browser drops it anyway?

PR #16, opened 2018-03-08
[#154558816] Add organisation templates

carolinegreen:
## What

Add template for showing organisations using the GOV.UK design system.  Use design system for title and fonts.

We may want a path for having no organisations and will probably want to look at this later.

First pass at creating new component for navigation panel on this page as it does not exist in the design system.  As this is specific to our app we will keep it here for now.  Roughly following the pattern of the design system for this component.  However some more work needs to be done on it to ensure that it's accessible and that the structure of it is reasonable.  It also needs a readme to describe more thoroughly what it is.  We might want to write tests around this as well.

Usually on frontend pull requests it's useful to put a before and after to see what the changes are.  There is no before for this page so putting in a screenshot of how it looks now.

screen shot 2018-03-08 at 15 18 07

## How to review

Code review, run the application, look at it.

## Who can review

Anyone but me, @paroxp, @camelpunch

### DIFF
--- a/src/layouts/govuk.njk
+++ b/src/layouts/govuk.njk
@@ -4,7 +4,7 @@
 
 {% set logo_link_title = 'Go to the GOV.UK PaaS Admin Tool Homepage' %}
 
-{% block page_title %}GDS Node.js boilerplate{% endblock %}
+{% block page_title %}PaaS Administration{% endblock %}
 
 {% block head %}
   <link href="./govuk.screen.scss" media="screen" rel="stylesheet" type="text/css" />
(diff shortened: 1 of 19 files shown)


src/layouts/govuk.njk
```diff
@@ -26,8 +26,13 @@
 {% block content %}
   <div class="govuk-o-width-container">
     <main class="govuk-o-main-wrapper">
+    <h1 class="govuk-heading-l">
+      {% block title %}
+      {% endblock %}
+    </h1>
       {% block main %}
         No content rendered to 'main' block
+        }
```
alext: Is this meant to be here. It looks unmatched to me.

alext:
The commit "Add run.sh to .gitignore" could do with some explanation as to why this is needed.

src/layouts/govuk.njk
carolinegreen: No, good spot.

PR #17, opened 2018-03-09
[#154558816] Organisation overview

paroxp:
## What

It wasn't clear what was wrong when it came to the request errors in
axios. We can capture that by try and catch and then throw a specific
error message from the request.

We cannot justify the use of synchronous approach express enforces. It
may be confusing to some people on the team as to why and when we use
async/await or promises.

We're creating a simple handler, that will take care of that for us.
It's a matter of including it when needed. Refactor to come later.

It's time to glue things together for spaces. We're adding some extra
functionality in order to download different pieces from API and put
them together in a new template parsed to end user.

We'd like to cover the spaces module as much as possible. We think these
should go in a separate commit in order to keep a bing chunk of code
separated.

There is no need for the layout to automatically include the h1 on every
page. Instead, we're moving it to the organisation page as it will be
the only one with this kind of layout.

## How to review

- Code review
- Run tests `npm test` (travis should do that)
- Run the application (You may want to use Rafal's environment. 
  No need to setup your own...)
  ```
  OAUTH_AUTHORIZATION_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/authorize" OAUTH_TOKEN_URL="https://uaa.rafalp.dev.cloudpipeline.digital/oauth/token" OAUTH_CLIENT_ID="rafal-login" OAUTH_CLIENT_SECRET="secret" SERVER_ROOT_URL="http://localhost:3000" API_URL=https://api.rafalp.dev.cloudpipeline.digital npm start
  ```
- Click through to single organisation
- Experience the overview of your organisation

Worth to mention, we compromised on some design details for different 
reasons. Happy to provide more details if interested.

## Who can review

Neither @chrisfarms nor myself.

### DIFF
--- a/src/app/sync-handler.js
+++ b/src/app/sync-handler.js
@@ -1,0 +1,7 @@
+import {internalServerErrorMiddleware} from '../errors';
+
+export default function syncHandler(f) {
+  return (req, res) => {
+    f(req, res).catch(err => internalServerErrorMiddleware(err, req, res));
+  };
+}
--- a/src/app/sync-handler.test.js
+++ b/src/app/sync-handler.test.js
@@ -1,0 +1,42 @@
+import {test} from 'tap';
+import express from 'express';
+import request from 'supertest';
+import pino from 'pino';
+import pinoMiddleware from 'express-pino-logger';
+import {internalServerErrorMiddleware} from '../errors';
+import syncHandler from './sync-handler';
+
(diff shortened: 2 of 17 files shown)


samcrang:
I experienced the overview and was impressed. 💯. Would experience again. 

There are however some conflicts that need merging.

paroxp:
Rebased. Sorry, we've merged something else in the mean time.

PR #26, opened 2018-03-26
Add sub navigation to all pages that deserve it

paroxp:
## What

Previously, we had to include the sub navigation on every page that
might have required it. The approach quickly proven we will forget about
certain places meaning the sub navigation will not be displayed.

We're now including the if statement in the main layout, checking
whether an `organization` variable has been provided from the controller
and decide on displaying the nav at all times.

## How to review

- Code sanity check
- Optionally, run the application using rafal's env and see if the subnav is displayed on almost every page you can click to except the org listing

### Before

screen shot 2018-03-27 at 12 07 35

### After

screen shot 2018-03-27 at 12 08 39

### DIFF
--- a/src/applications/overview.njk
+++ b/src/applications/overview.njk
@@ -6,11 +6,7 @@
   {{ application.entity.name }} - Application Overview
 {% endblock %}
 
-{% block main_classes %}with-subnav{% endblock %}
-
 {% block main %}
-  {% include "../organizations/subnav/element.njk" %}
-
   <div class="govuk-o-grid">
     <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
       {{ govukBackLink({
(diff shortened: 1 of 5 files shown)


carolinegreen:
Could you add in before and after pictures?  Realised how much that helps with context and just put it in the small pull request I opened as well.

paroxp:
Added. Not sure if all that useful. Still need to spin up the application and click around to see if it's working.

carolinegreen:
Thanks, sorry didn't see this earlier.  Still going to spin up the app, but it makes it easier to know where to look for visual design changes quickly and where to expect them.

carolinegreen:
Looks good to apart from one thing.  I think this must have gone through in a separate story as it's not in the code, but seeing it now.  The heavy line under org overview doesn't change when you switch to team members, and I think it's supposed to be a visual indicator of where you are.  So either worth fixing it, or just removing it for this story and create another story to fix it later.

carolinegreen:
Going to merge this so it goes through, and we can add a story about the lines later on.

PR #27, opened 2018-03-27
Add navigation

carolinegreen:
Add basic navigation for PAAS admin.  We need this to navigate between pages at a top level.  This is part of the main govuk template but should probably be separated out into a separate template later on.

Todo
-  Make navigation mobile friendly
-  Indicate which section user is on
-  Connect routing to templates rather than having urls hardcoded.  I think this needs to be addressed in another story as we are about to rewrite our routing completely.

Before:

screen shot 2018-03-27 at 11 24 07

After:

screen shot 2018-03-27 at 11 27 54

This can be reviewed by running the application and making sure the links are there.  The only link that links to anything currently is organisations as I don't think the other sections yet exist.

Anyone can review this who isn't me.

### DIFF
--- a/src/layouts/_header.scss
+++ b/src/layouts/_header.scss
@@ -1,0 +1,27 @@
+.navigation {
+  float:right;
+  display: block;
+  margin-right: $govuk-spacing-scale-3;
+  margin-top: 7px;
+  padding-top: 0;
+  @include govuk-font-bold-16;
+
+  a {
+    color: $govuk-white;
+    text-decoration: none;
+
+    &:hover {
+     text-decoration: underline;
+    }
+  }
+
+  ul {
+    margin: 0;
+  }
+
+  li {
+    list-style-type: none;
+    display: inline;
+    margin-left: $govuk-spacing-scale-3;
+  }
+}
--- a/src/layouts/govuk.njk
+++ b/src/layouts/govuk.njk
@@ -11,6 +11,17 @@
   <link href="./govuk.screen.scss" media="screen" rel="stylesheet" type="text/css" />
 {% endblock %}
 
+{% block inside_header %}
+  <div class="navigation">
+    <ul>
+      <li><a href="/organisations">My Organisations</a></li>
+      <li><a href="https://docs.cloud.service.gov.uk/">Documentation</a></li>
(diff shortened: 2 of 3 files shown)


src/layouts/govuk.njk
```diff
@@ -5,6 +5,8 @@
 
 {% set logo_link_title = 'Go to the GOV.UK PaaS Admin Tool Homepage' %}
 
+{% block inside_header %} <div class="navigation"><ul><li><a href="/organisations">My Organisations</a></li><li><a href="#">Documentation</a></li><li><a href="#">Support</a></li><li><a href="#">Sign out</a></li><ul>{% endblock %}
```
paroxp: Could we make each `li` a new line? The links do not need to be hash tags :)

Documentation: `https://docs.cloud.service.gov.uk/`
Support: `mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk`
Sign out: `/auth/logout`

src/layouts/_header.scss
```diff
@@ -0,0 +1,27 @@
...
+  ul {
+    margin: 0px;
+  }
+
+  li {
+    list-style-type: none;
+    display: inline;
+    margin-left: 15px;
+  }
+}
```
paroxp: missing blank line :trollface:

src/layouts/_header.scss
```diff
@@ -0,0 +1,27 @@
+.navigation {
+  float:right;
+  display: block;
+  margin-right: 15px;
```
paroxp: You're hardcoding some `px` in few places. Would we benefit from the use of the variables like `$govuk-spacing-scale-3`?

carolinegreen: Was trying to do it fast as thought we needed it out yesterday for the deploy story, but if there's more scope can go and find the mixins.

src/layouts/govuk.njk
carolinegreen: Yeah of course, this was to have something out for deployment, and thought we had less time.

src/layouts/_header.scss
carolinegreen: Yeah, honestly need to add it to my editor.

PR #29, opened 2018-04-03
[#156101804] Implement router

paroxp:
## What

The idea was, we would stop using the express default router due to some issues we have with it. This seemed like the perfect opportunity to use TypeScript to save us some problems with misusing the router.

New router, means we'd need to readapt the handlers we have. Whilst refactoring anyway, I thought it would not harm us to convert everything over to TypeScript. Some of the bits that were left in JS were just standing out. And in order to fully benefit from types, it was natural to convert it all.

~~My only regret is the:~~

```
  options: {
    transpileOnly: true
  }
```

~~in webpack config. At some point I'll be aiming to take it out. ;/~~

I have no regrets... I faced all the issues it had. It previously caused me delay in attempt to discover a misspelled variable in a file I have rarely opened. That's exactly the reason we introduced TypeScript in the first place, therefore it will now error at all times.

## How to review

- all the review
- run tests (done with travis)
- run application (use rafal's environment)
- navigate around - see if everything is still in place
- attempt to create a custom route, see how it feels
- comment on everything! we need to be happy with it


src/organizations/organizations.ts
```diff
@@ -0,0 +1,13 @@
+import { IResponse, IStringToStringMap } from '../lib/router';
+
+import { IContext } from '../app/context';
+
+import organizationsTemplate from './organizations.njk';
+
+export async function listOrganizations(ctx: IContext, _params: IStringToStringMap): Promise<IResponse> {
```
chrisfarms: I think `IStringToStringMap` would be better named something like `IParams` ... I don't think there's much value in reusing these types of types

chrisfarms:
NOREGRATS

PR #34, opened 2018-04-19
[#156761419] Only amend permissions on change

paroxp:
## What

We were making a lot of calls each time a single permission would change.
Meaning we have likely hit the problem with API responding with errors
when "unsetting" permission that has not been previously set.

We decided to record current and desired permission in HTML for later
comparison in the handlers and preparing the call that matters to end
user.

## How to review

1. Sign in as a user with only OrgManager role
1. Attempt to update the roles of another user
1. Be successful

OR 

1. Sign in as a user with only OrgManager role
1. Attempt to invite a user
1. Be successful

## Who can review

Neither @dcarley nor myself.

### DIFF
--- a/src/users/users.test.ts
+++ b/src/users/users.test.ts
@@ -14,69 +14,6 @@ import { IContext } from '../app/context';
 
 const tokenKey = 'secret';
 
-// tslint:disable:max-line-length
-nock('https://example.com/api').persist()
-  .get('/v2/organizations/3deb9f04-b449-4f94-b3dd-c73cefe5b275').reply(200, cfData.organization)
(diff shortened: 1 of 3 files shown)


src/users/users.test.ts
```diff
@@ -102,286 +39,584 @@ const ctx: IContext = {
...
+    },
+  };
+
+  return {
+    ...defaultRoles,
+    ...setup,
+  };
+}
+
+test('ordinary set of tests', async _t => {
```
chrisfarms: Is this an odd title?

paroxp: It is xD Naming is hard and I couldn't come up with a better title for "current and old style of tests, which work but are somewhat ugly"

PR #35, opened 2018-04-19
[#156761419] Less verbose logs

paroxp:
## What

Pino by default would provide a lot of information through to the logs.
This may be helpful at some point, but in practice is annoying and
dangerous as a session cookie is printed.

We have a habbit of copying the error log and pasting in places, meaning
it could potentially leak the session. Let's avoid it by serializing the
request and response for the logger to use.

## How to review

- Run application
- Login to the app
- Cause an error, like visit `/organisations/fake` route
- Observe logs - experience lack of headers in the logs

## Who can review

Neither @dcarley nor myself

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -4,6 +4,7 @@ import express from 'express';
 import pinoMiddleware from 'express-pino-logger';
 import staticGzip from 'express-static-gzip';
 import helmet from 'helmet';
+import { IncomingMessage, ServerResponse } from 'http';
 import { BaseLogger } from 'pino';
 
 import auth from '../auth';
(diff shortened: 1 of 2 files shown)


src/app/app.ts
```diff
@@ -33,7 +34,18 @@ export interface IAppConfig {
...
 
-  app.use(pinoMiddleware({logger: config.logger}));
+  app.use(pinoMiddleware({
+    logger: config.logger,
+    serializers: {
+      req: (req: IncomingMessage) => ({
+        method: req.method,
+        url: req.url,
+      }),
+      res: /* istanbul ignore next */ (res: ServerResponse) => ({
```
chrisfarms: I think coverage could be caught for this using the "main" tests, but I don't feel strongly enough to request a change for this.

paroxp: It felt bad when I put that in... It's a simple check, yet feels unnecessary...

PR #38, opened 2018-04-25
Update space sections in org overview

carolinegreen:
## What

Changed the structure from divs to sections with a structure inside of headings, paragraphs and lists as this makes more sense semantically than divs.  Also added thinner border as a design change update.  Changed some of the css so that the mobile view also matches the protoype.  The entire box is a link when the user hovers over.  We will need to look at how accessible this pattern is, and ways of making it more accessible.

Before:
screen shot 2018-04-25 at 18 50 05

After:
screen shot 2018-04-25 at 18 50 42

Mobile view Before:

screen shot 2018-04-25 at 18 53 30

Mobile view After:

screen shot 2018-04-25 at 18 50 57

## How to review

Run it, check it against Steve's design and make sure it works.

## Who can review

Anyone who isn't me

### DIFF
--- a/src/spaces/spaces.njk
+++ b/src/spaces/spaces.njk
@@ -70,28 +70,29 @@
           </a>
         </h3>
       </div>
-      
+
       <div class="govuk-o-grid__item govuk-o-grid__item--one-third govuk-o-grid">
         <div class="govuk-o-grid__item govuk-o-grid__item--full">
           The space has {{ space.entity.apps | length }} apps:
         </div>
 
-        <div class="govuk-o-grid__item govuk-o-grid__item--one-half running">
-          {{ space.entity.running_apps | length }} running
+      <div class="space-information">
+        <div class="govuk-o-grid__item govuk-o-grid__item--one-third govuk-o-grid">
+          <h4 class="govuk-o-grid__item govuk-o-grid__item--full">
+            The space has {{ space.entity.apps | length }} apps:
+          </h4>
+          <ul>
+            <li class="running">
+              {{ space.entity.running_apps | length }} running
+            </li>
+            <li class="stopped">
+              {{ space.entity.stopped_apps | length }} stopped
+            </li>
+          </ul>
         </div>
 
-        <div class="govuk-o-grid__item govuk-o-grid__item--one-half stopped">
-          {{ space.entity.stopped_apps | length }} stopped
-        </div>
-      </div>
-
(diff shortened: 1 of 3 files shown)


src/spaces/spaces.njk
```diff
@@ -57,50 +57,51 @@
...
   {% for space in spaces %}
-    <div class="govuk-o-grid space">
-      <div class="govuk-o-grid__item govuk-o-grid__item--one-third">
-        <h3 class="govuk-heading-m">
-          <span class="govuk-caption-m">
-            Space:
-          </span>
-
-          <a href="{{ linkTo('admin.organizations.spaces.view', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid}) }}" id="space-{{ space.metadata.guid }}" class="govuk-link govuk-!-w-bold">
+    <a href="{{ linkTo('admin.organizations.spaces.view', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid}) }}" id="space-{{ space.metadata.guid }}">
```
paroxp: Can we not wrap the entire section in a link? I realise it's how it's done in the design, but I have some flashing lights in my head which I cannot really explain ;/
It has only became OK with modern browsers and HTML5, which I think may mean problems with some readers, parsers or renderers.

Good example is Firefox, it selects additional parts of the page on `:focus`. I'm sure it could be solved with `a { display: block; }` or something, but I'm wary of possible edge cases.

I'd rather it be done with the use of JS if needed. `if box clicked, find link and execute`. But that's out of scope.

I'm not going to fight for it though... What do you think? /cc @carolinegreen @chrisfarms

carolinegreen: This is kind of the problem I was mentioning.  It's not the most accessible pattern either.  But the problem is that if it's implemented the way it was before, it's not necessarily obvious that the user can go to the next section with one small link.  We could really do with testing whether people would know to go to the next page.

carolinegreen:
@paroxp I had the spacing in another story.

src/spaces/spaces.njk
paroxp: You're right. But aren't we trying to fix the problem that doesn't exist yet? We didn't hear problem about this yet from the people who have tested it.

Whilst you know it's a problem for readers and while I use firefox, it does have weird behaviour 🤔

carolinegreen: I've seen this happen on other interfaces with small links like this.  It doesn't make sense that there's a tiny link and that it goes somewhere, with the way it was implemented and we never discussed it with design and content.

PR #41, opened 2018-05-02
[#156916095] Generating monthly usage reports

chrisfarms:
## What

We have moved the basic UI for rendering usage reports from paas-billing to paas-admin so that we can use it to generate bills. To do this we implemented a BillingClient to talk to our billing API, added a new handler for fetching/grouping the events for an org by "name" and sketched out rendering the data table.

* **This is NOT the final design of the report**: a future story will implement the new designs that have been created, it is intended for internal use for now.
* **This is NOT currently exposed to all tenants**: there is no "billing" link in the UI (you must know the url to visit it) and paas-billing currently restricts access to tokens with `admin`, `admin_read_only` or `global_auditor` scope. A future story will expose this feature for all tenants with BillingManager and/or OrgManager roles.

## How to review

* Code review
* Run the application against a dev environment that has some billing-db data.
* Visit `/organisations/{org_guid}/statements/{YYYY-MM-DD}`
* Check that things make sense

## Who can review

Not @chrisfarms or @paroxp


src/lib/billing/billing.test.ts
```diff
@@ -0,0 +1,117 @@
+import nock from 'nock';
+import { test } from 'tap';
+
+import { config } from '../../app/app.test';
+
+import { BillingClient } from '.';
+
+// tslint:disable:max-line-length
+nock(config.billingAPI)
+  .get('/billable_events?range_start=2018-01-01&range_stop=2018-01-02&org_guid=org-guid-500-no-data').reply(500)
```
dcarley: Could these individual mocks be moved into the specific tests that they relate to so that it's easier to understand each test?

src/app/router.ts
```diff
@@ -72,7 +72,7 @@ const router = new Router([
   {
     action: statement.viewStatement,
     name: 'admin.statement.view',
-    path: '/organisations/:organizationGUID/statements',
+    path: '/organisations/:organizationGUID/statements/:rangeStart',
```
dcarley: It took me a while to figure out the behaviour, where:

- `2018-05-01` gave today only with costs
- `2018-05-02` gave tomorrow only with no costs
- `2018-04-01` gave last month in full
- `2018-04-15` gave a partial month

Could this take a month rather than a start date and always report on the calendar month?

package.json
```diff
@@ -32,11 +32,12 @@
     "@types/express": "4.11.1",
     "@types/helmet": "0.0.37",
     "@types/jsonwebtoken": "7.2.6",
+    "@types/moment": "2.13.0",
```
dcarley: Can this commit be squashed with the commit that uses it so that it's easier to understand how/why it's used?

src/lib/billing/billing.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    const data: ReadonlyArray<IBillableEventResponse> = response.data;
+
+    return data.map(parseBillableEvent);
+  }
+}
+
+function parseTimestamp(s: string): Date {
+  return new Date(s); // FIXME: Date parsing is unreliable consider momentjs
```
dcarley: Is this FIXME still relevant after adding momentjs as a dependency?

README.md
```diff
@@ -69,12 +69,11 @@ npm test
 Start the server in development mode
 
 ```sh
-OAUTH_AUTHORIZATION_URL="https://login.$DEPLOY_ENV.dev.cloudpipeline.digital/oauth/authorize" \
-OAUTH_TOKEN_URL="https://uaa.$DEPLOY_ENV.dev.cloudpipeline.digital/oauth/token" \
```
dcarley: It seems like this commit could be squashed with `Create clients when needed` which removed the use of these variables.

There's also a reference to these environment variable names in `src/main.test.ts`.

src/users/users.ts
```diff
@@ -392,12 +389,19 @@ export async function editUser(ctx: IContext, params: IParameters): Promise<IRes
...
   const cf = new CloudFoundryClient({
     accessToken: ctx.token.accessToken,
     apiEndpoint: ctx.app.cloudFoundryAPI,
   });
 
+  const isAdmin = ctx.token.hasScope(CLOUD_CONTROLLER_ADMIN);
+  const isManager = await cf.hasOrganizationRole(params.organizationGUID, ctx.token.userID, 'org_manager');
+
+  /* istanbul ignore next */
+  if (!isAdmin && !isManager) {
```
dcarley: Some of the changes in this commit don't seem related to signingKeys refactoring. Could you update the commit message to describe what it's doing? It feels a bit odd to partially refactor the signingKeys stuff here prior to the other story being worked.

paroxp: You've managed to make me doubt at what we're doing. I've reworded that commit entirely as I got confused at what we're trying to accomplish. See if that's better?

src/statement/statement.ts
```diff
@@ -0,0 +1,108 @@
...
+    event.price.details.forEach(detail => {
+      if (detail.currencyCode === 'USD') {
+        usdExchangeRate = detail.currencyRate;
+      }
+    });
+
+    if (!resource) {
+      return {...rest, [key]: {
+        ...event,
+        planName: event.price.details.map(pc => pc.planName.replace('Free', 'micro'))
```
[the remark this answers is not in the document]
chrisfarms: In the "propper" design of this page we have widget for expanding each pricing row to show the "details" so this hacky "grab the first exchange rate you find from the first details" thing can go away.

But yes I'll leave this as a reminder for kickoff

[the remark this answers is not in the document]
chrisfarms: note added about removing plan name fudging 👍

src/app/router.ts
dcarley: The new error message doesn't make it to the browser, but hopefully that'll be fixed in https://www.pivotaltracker.com/story/show/156760174

PR #44, opened 2018-05-03
Introduce the use of /v2/info

paroxp:
## What

Currently, we have to provide a lot of environment variables before
running the application. This is annoying some of these could be sliced
out.

We're going to request `/v2/info` and set those up in the application
itself, to reduce devops stuff.

## How to review

- Make sure the tests pass
- Run the application
- Make sure you can log in
- You can check if the log on start shows correct urls.

## Who can review

Neither @chrisfarms nor myself.


keymon:
> to reduce devops stuff

:cry: 

https://en.wikipedia.org/wiki/DevOps

keymon:
Additionally to enable "Manual operation overhead", it is claimed that this PR would parse the two endpoints for authentication and tokens, which are different in our setup: `http://login.<system>` and `http://uaa.<system>`.

In our previous deployments we were setting `UAA_URL` as `https://uaa<system>`, which caused the IDP  federation fail, as the oauth2 clients were configured to work on the `https://login.<system>` domain. Because that, we were getting the following error when authenticating with google:

```
Error: redirect_uri_mismatch

The redirect URI in the request, https://uaa.cloud.service.gov.uk/login/callback/google, does not match the ones authorized for the OAuth client. To update the authorized redirect URIs, visit: https://console.developers.google.com/apis/credentials/oauthclient/.......apps.google
```

Possibly the reviewer of this PR should double check that this functionality would work with the new configuration from `/v2/info`

keymon:
Related: https://github.com/alphagov/paas-cf/pull/1346

src/main.ts
```diff
@@ -31,16 +32,50 @@ function onShutdown() {
   process.exit(0);
 }
 
-async function main(cfg: IAppConfig) {
-  const server = new Server(app(cfg), {
+async function main() {
+  const cloudFoundryAPI = expectEnvVariable('API_URL');
+  let authorizationURL = process.env.AUTHORIZATION_URL;
+  let uaaAPI = process.env.UAA_URL;
```
dcarley: Can you describe why these would be overridden by environment variables?

It'd be simpler (easier to understand the code and any possible misconfigurations) if they weren't.

src/app/app.test.ts
```diff
@@ -25,6 +25,7 @@ export const config: IAppConfig = {
   oauthClientSecret: 'secret',
   cloudFoundryAPI: 'https://example.com/api',
   uaaAPI: 'https://example.com/uaa',
```
dcarley: Can we rename all instances of `uaaAPI` to `tokenAPI` so that it's consistent with the naming from `/v2/info` and easier to differentiate between the two endpoints (because we've already been confused by the `redirect_uri` problem)?

paroxp:
We took the comments offline. I've applied some changes to find a common ground.

PR #46, opened 2018-05-11
[#154329817] Implement pricing calculator

paroxp:
## What

We're implementing the pricing calculator for our tenants to have an idea
of the costs of using PaaS.

This involves extending the billing client to be more consistent with the
API.

**Bonus:**

TIL, including single const from inside another test file would trigger
the tests in that file to be run... This is unnecessary as it would take
time to run them and add some fake numbers towards our end test count.

We've found previously introduced HTML error where the closing tag is
inconsistent with the opening one.

## How to review

- Code review
- Run tests (done with travis)
- Run application
- Navigate to `/calculator`
- Ignore inconsistency with the design
- Have a play around

## Who can review


LeePorte:
Some of the plan data is a little confusing, but this will be cleaned up with the upcoming plan rename within the billing app.

PR #50, opened 2018-05-18
[#156431805] Add tabs UI for applications/services

chrisfarms:
## What

We implemented some tabs

## How to review

* Code review
* Run it, click around
* Look at it

## Before

screenshot 2018-05-18 15 48 03

## After

screenshot 2018-05-18 15 50 31

## Who can review

Not @carolinegreen or @chrisfarms

### DIFF
--- a/src/spaces/backing-services.njk
+++ b/src/spaces/backing-services.njk
@@ -1,0 +1,51 @@
+{% extends "../layouts/govuk.njk" %}
+
+{% block page_title %}
+  {{ space.entity.name }} - Overview
+{% endblock %}
+
+{% block main %}
+  <h1 class="govuk-heading-l">
+    <span class="govuk-caption-l">Space</span>
+    {{ space.entity.name }}
+  </h1>
+
+  {% include "./_tabs.njk" %}
+
+  {% if space.entity.services.length > 0 %}
+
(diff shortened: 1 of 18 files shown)


src/spaces/backing-services.njk
```diff
@@ -0,0 +1,51 @@
...
+          <th class="govuk-c-table__header" scope="col">Type</th>
+          <th class="govuk-c-table__header" scope="col">Plan</th>
+          <th class="govuk-c-table__header" scope="col">State</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-c-table__body">
+        {% for service in space.entity.services %}
+          <tr class="govuk-c-table__row">
+            <th class="govuk-c-table__header" scope="row">
+              <a href="{{ linkTo('admin.organizations.spaces.services.view', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.guid}) }}" class="govuk-link">{{ service.name }}</a>
```
LeePorte: This as a link doesn't take you anywhere, should it?

chrisfarms: Yep it should ... good catch.

I have added a commit to fix it and added a test to catch duplicated names used for routes (which was the cause of the issue)

PR #53, opened 2018-05-22
Fix font sizes, whitespace and add copy.

carolinegreen:
## What

Fix font sizes, whitespace, and emphasis on what is bold in the table on the member's page to be closer to Steve's design.  Also add copy describing user roles.

## Before

screen shot 2018-05-22 at 16 35 06

screen shot 2018-05-22 at 16 24 47

## After

screen shot 2018-05-22 at 16 22 44

screen shot 2018-05-22 at 16 22 59

## How to review

Run the application.  Check design against @stephenjoe1's prototype.  Check copy.

## Who can review

Anyone who isn't me.

### DIFF
--- a/src/users/users.njk
+++ b/src/users/users.njk
@@ -13,7 +13,7 @@
       Team members
     </h1>
 
-    <p class="govuk-body-l">
+    <p class="govuk-body-m">
       Members can have organisation and space level roles.
       <br />
       To edit a members role go to their profile.
@@ -23,7 +23,7 @@
   <div class="govuk-grid-column-one-third text-right">
     {% if isAdmin or isManager %}
       {{ govukButton({
-        text: "Invite a new team member",
+        text: "Invite a new member",
         href: linkTo('admin.organizations.users.invite', {organizationGUID: organization.metadata.guid})
       }) }}
     {% endif %}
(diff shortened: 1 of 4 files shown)


src/users/users.njk
```diff
@@ -6,19 +6,14 @@
...
-  {{ super() }}
-  <link rel="stylesheet" type="text/css" href="./users.scss">
-{% endblock %}
-
 {% block main %}
 <div class="govuk-o-grid">
   <div class="govuk-o-grid__item govuk-o-grid__item--two-thirds">
-    <h1 class="heading-large">
+    <h1 class="govuk-heading-l">
       Team members
```
paroxp: You took the `Team` away from the button. Do you think it's worth taking away from here too?

carolinegreen: I was leaving "team" here as I thought it was nice as a an introduction and we could shorten it to "members" everywhere else as people already had the context that these were "team members" from the heading.  Putting "team" in front of the heading also makes it seem a little bit less repetitive when the paragraph below starts with "members".

carolinegreen: Maybe we should have content designer look at it though.

paroxp: Maybe, my only worry is that introduction of a new unfamiliar to CF word which may confuse people. What is a team? We already have Organisation Members and Space Members etc.

PR #54, opened 2018-05-30
[#157730702] Use npm audit

chrisfarms:
## What

Swap out snyk for npm audit for checking for issues with dependencies during `npm test` runs.

## How to review

* Check that the new `npm run test:audit` returns a full audit
* Check that the `npm run test:vulnerabilities` looks like it will return a non-zero exit code when a high or critical issue is found

## Who can review

Not @chrisfarms


package.json
```diff
@@ -72,7 +73,7 @@
     "tslint-microsoft-contrib": "5.0.3",
     "typescript": "2.8.1",
     "url-loader": "1.0.1",
-    "webpack": "4.4.1",
+    "webpack": "^4.10.2",
```
paroxp: Can we pin the versions? :D

paroxp:
Worth to mention, `npm install` in order.

package.json
chrisfarms: well the lock file kinda does this already ... but ok, one sec :)

chrisfarms: done ... I intentionally left the `npm` only pinned to ^6.

PR #60, opened 2018-06-08
[#157278274] Allow user deletion

paroxp:
## What

We'd like to allow our tenants to remove certain users from their
orgranisation. This can be done via the secret `users` role being unset
on already existing org user. We're adding views for delete
confirmation and the success page.

## How to review

- Run the application
- Attempt to delete user from one of the organisations


carolinegreen:
I think it would be worth adding screenshots.  I'm also wondering if it would make more sense to use buttons like these:

http://alphagov.github.io/digitalmarketplace-frontend-toolkit/button.html

I realise the design has links, but that the user journey has also changed a bit since then.   Also removing a user is an action, so having a button like saving role changes is currently would be more uniform.  Probably worth talking with @stephenjoe1 too.

paroxp:
Didn't add screenshots, as there is no before or after. It has no designs either aside from the remove link.

> Also removing a user is an action, so having a button like saving role changes is currently would be more uniform

The initial `link` is still a link IMO. It redirects user to a confirmation page, where you have to confirm/perform the actual action represented by a button.

carolinegreen:
You’re right about it being a link, but it's a bit of an unusual case because it's leading to a page that is only a confirmation page and nothing else, so in a way the link to it is part of the action.  This is why I'm wondering if a link represents the meaning in the best way in this particular situation.  In other teams in the past there have been various approaches to this, including using something in between a link and a button, or confirmation flash messages.  The way it currently looks just now makes it appear as if changing a user role is a more drastic action than removing the user with the bright green button right next to a more understated link.

I think screenshots can be useful if you are adding a new user journey, even if there’s not a before as it makes it easier to see what’s there visually quite quickly.

PR #62, opened 2018-06-14
[#156760174] Stop users from removing the last org manager

carolinegreen:
https://www.pivotaltracker.com/n/projects/1275640/stories/156760174

## What

Prevent users from removing the last org manager as this means they remove the ability to manager organisations from themselves, which if enable results in an error.  As a follow up it's worth adding copy that we recommend having at least two organisation managers at the very least.

## Disabled for removing the last org manager.

screen shot 2018-06-14 at 17 46 26

## How to review

Run the application as a user below admin level, and try to remove all the org managers.

## Who can review

Anyone who isn't me.


keymon:
I rebased this

PR #63, opened 2018-06-15
[#157510099] Fix sticky and remove stray debug in pricing calculator

carolinegreen:
## What

The sticky effect on the summary box for the calculator became broken somewhere along the way and so this is fixed.  It currently isn't visible as the list of services isn't currently long enough for sticky to happen, but if the list of services does become longer this design won't be lost.

Also removed a stray parameter dump as this would have been for debugging purposes.

## How to review

Run the application.  Make sure everything still works.

## Who can review

Anyone who isn't me.

### DIFF
--- a/src/calculator/calculator.njk
+++ b/src/calculator/calculator.njk
@@ -22,86 +22,84 @@
 
 {% block main %}
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-one-half">
+    <div class="govuk-grid-column-two-thirds">
       <h1 class="paas-heading-l">Estimate your monthly costs</h1>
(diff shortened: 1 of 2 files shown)


keymon:
Is this related to any story in pivota?

keymon:
Apparently the most related story is: https://www.pivotaltracker.com/story/show/157510099

I will tag it as such

paroxp:
The fix here has been accidentally applied somewhere else. But the PR points the data dump on the page, which makes it somewhat needed.

I'm going to close it down and spin up a new one removing the datadump.

Bear with me @keymon

PR #67, opened 2018-06-18
Let's not dump the data

paroxp:
## What

Although this page is not available to anyone except for the admins,
we'd like to remove the dump just in case we'd forget it in the future.

## How to review

- Code review and sanity check

## Who can review


paroxp:
I'm going to merge it myself. Small enough change that doesn't interact with functionality of the app.

PR #68, opened 2018-06-18
[#157185422] Fix how the space quota is displayed

keymon:
What
----

When displaying the space quota in the list of spaces we
check if there is a space quota to show the space limit. But we
incorrectly refer to the wrong valiable, as the quota attribute
of the space has a `entity` field.

This causes the interface to show a NaNgb from a null/1024 operation.

We update the interface to use the right value.

How to review
----

Reproduce this problem creating some space quota:

```
cf create-org foo
cf target -o foo
cf create-space dev
cf target -s dev
cf create-space-quota foo -m 1gb
cf set-space-quota dev foo
```

You can run this code from your dev environment like:

Provided you have logged in CF and have AWS credentials

1. Update uaa user:

```
DEPLOY_ENV=...
uaac target https://uaa.${DEPLOY_ENV}.dev.cloudpipeline.digital
uaac token client get admin -s $(aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/cf-secrets.yml  - | grep secrets_uaa_admin_client_secret | cut -f2 -d " ")
uaac client update paas-admin --redirect-uri http://localhost:3000/auth/login/callback
```

2. Get the command to start the server:

```
cf target -o admin -s public
cf env paas-admin | awk '/User-Provided/ { printing=1; next} /^$/ { printing=0 ; next} printing  {gsub(": ","=\""); gsub("$", "\" \\"); print; next  } END { print "npm start"}'

```

3. Undo the paas-admin client change once done:

```
uaac client update paas-admin --redirect-uri https://admin.${DEPLOY_ENV}.dev.cloudpipeline.digital/auth/login/callback
```

Check that the new code displays the right units.

Who can review
----

Anyone but @keymon

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,15 @@
-## What
+What
+----
 
 Describe what you have changed and why.
 
-## How to review
+How to review
+-------------
 
 Describe the steps required to test the changes.
 
-## Who can review
+Who can review
+---------------
 
 Describe who can review the changes. Or more importantly, list the people
 that can't review, because they worked on it.
--- a/README.md
+++ b/README.md
@@ -87,6 +87,35 @@ gorouter's self-signed cert (e.g. on bosh-lite).
 You should be able to edit files in the `./src` directory and the changes will
 automatically be updated.
 
+## Quick local start for review/troubleshooting
+
+Provided you have logged in CF and have AWS credentials
+
+1. Update uaa user:
+
+```
+DEPLOY_ENV=...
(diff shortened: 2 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,15 @@
-## What
+What
```
chrisfarms: Not a fan of the nice # syntax?

keymon: No, see commit: https://github.com/alphagov/paas-admin/pull/68/commits/9347483d3f010c13c2bbaae030867f4425e1d571

PR #71, opened 2018-06-19
[#154590310] Billing for tenants

paroxp:
## What

We'd like to allow our tenants to view their own billing statements.

Most of the functionality is there. We're adding few checkpoints to see if the user attempting to display the data is indeed authorised to do so. We're making it easier for them to find the data and more friendly approach to display different states of it.

We're also adding a magic route that takes query parameter and uses it to redirect to correct billing statement. If none provided, it should redirect to current month's.

## How to review

- Run the application
- Log in as billing or org manager
- Attempt to find yourself around the billing

## Who can review

Neither @keymon nor myself.

### DIFF
--- a/src/statement/statement.njk
+++ b/src/statement/statement.njk
@@ -5,6 +5,7 @@
 {% from "@govuk-frontend/frontend/components/back-link/macro.njk" import govukBackLink %}
 {% from "@govuk-frontend/frontend/components/details/macro.njk" import govukDetails %}
 {% from "@govuk-frontend/frontend/components/table/macro.njk" import govukTable %}
+{% from "@govuk-frontend/frontend/components/select/macro.njk" import govukSelect %}
 
 {% block page_title %}
   Statement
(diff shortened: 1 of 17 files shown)


src/statement/statement.njk
```diff
@@ -43,66 +54,72 @@
...
-        <th class="govuk-table__header" scope="col" style="text-align: right">Ex VAT</th>
-        <th class="govuk-table__header" scope="col" style="text-align: right">Inc VAT</th>
-      </tr>
-    </thead>
-    <tbody class="govuk-table__body">
-      {% for guid, items in items | groupby('spaceGUID') %}
+  {% if items | length %}
+    {% if isCurrentMonth %}
+      <p class="paas-table-notification">
+        This statement show up to date provitional usage data for current month.
```
paroxp: `s/show/shows`

carolinegreen: Yeah, also provisional.  Should turn on your spellcheck.

PR #72, opened 2018-06-19
[#158454773] Display resend link only when applicable

paroxp:
## What

We'd like to allow org managers to resend the invitation link, only when
the user is not active in the UAA.

It does bring an extra call that slows the rendering of the view
slightly, but is improvement for the user experience.

Also:
It may be sending the wrong message when we claim to remove the user,
yet all we do is unfriend them from our organisation. Making it more
clear.

## How to review

- Run the application
- As org manager have a look around users
- Attempt to find inactive user and see they have the resend button applied

### DIFF
--- a/src/users/edit.njk
+++ b/src/users/edit.njk
@@ -60,7 +60,7 @@
     </form>
 
     <a href="{{ linkTo('admin.organizations.users.delete', {organizationGUID: organization.metadata.guid, userGUID: user.metadata.guid}) }}" class="govuk-link">
-      Remove user
+      Remove user from Organisation
     </a>
   </div>
 </div>
--- a/src/users/edit.njk
+++ b/src/users/edit.njk
@@ -45,11 +45,14 @@
       }) }}
     {% endif %}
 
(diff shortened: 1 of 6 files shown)


src/users/edit.njk
```diff
@@ -45,11 +45,14 @@
...
     {% endif %}
 
-    <form method="post" action="{{ linkTo('admin.organizations.users.invite.resend', {organizationGUID: organization.metadata.guid, userGUID: user.metadata.guid}) }}" class="govuk-!-mt-r6">
-      {{ govukButton({
-        text: "Resend user invite"
-      }) }}
-    </form>
+    {% if not isActive %}
+      <form method="post" action="{{ linkTo('admin.organizations.users.invite.resend', {organizationGUID: organization.metadata.guid, userGUID: user.metadata.guid}) }}" class="govuk-!-mt-r6">
+        <p>It'd appear the user did not setup their account yet.</p>
```
carolinegreen: Think this should at least be "it would."  The content designers have new rules about never using shortened versions of things as they found that it confused some people.

paroxp: Cool. Sounds good. Will change.

keymon: Personally I do not really get that sentence "It would appear". But I am not a native speaker.

Would it better something like "Apparently the user has not activated the account yet."?

PR #74, opened 2018-06-26
Update content

carolinegreen:
What
----

I've updated the content as much as possible from Paola's spreadsheet https://docs.google.com/spreadsheets/d/1oso5C_I1eR98Y0Hzb8hroAy3iIak1mnWj3spBF5B0_w/edit?usp=sharing  I haven't used the content in places where either, there was already new content,  the content's wording didn't make sense to me, or I wasn't sure if the new content was actually better than the old content.

How to review
-------------

I've written descriptions for the changes I've made in each commit following each row of the spreadsheet to make it easier to review.  It's probably also a good idea to run the application and go through the pages as you are looking at these content changes as a lot of the content relies on the context of the page, or which part of the user journey it is in.  For example something that is part of a table is surrounded by context that might not be obvious when just looking at plain text, and something deeper in the application has more context around it than something on the first page.

It would also be good if several different people reviewed this, so that we could get as many different views on this as possible as someone who is very familiar with Cloudfoundry is going to have very different ideas to what something means compared to someone who hasn't used it.

Who can review
---------------

Anyone who isn't me, but a few people please.

### DIFF
--- a/src/users/permissions.njk
+++ b/src/users/permissions.njk
@@ -2,20 +2,20 @@
 
 <h2 class="govuk-heading-l">Set org and space roles</h2>
 
-<h4 class="govuk-heading-s">Organisation roles</h4>
+<h4 class="govuk-heading-s">Organisation level roles</h4>
 
 <details class="govuk-details" role="group">
   <summary class="govuk-details__summary" role="button" aria-controls="details-content-0" aria-expanded="false">
     <span class="govuk-details__summary-text">
-      Org level role descriptions
+      What can these roles do?
     </span>
   </summary>
 
   <div class="govuk-details__text" aria-hidden="true">
-    <h3 class="heading-small">Org billing manager</h3>
-    <p>An Org billing manager is a point of contact for billing</p>
     <h3 class="heading-small">Org manager</h3>
     <p>An Org manager can create/delete spaces and edit user roles. The Org Managers would typically be senior staff: for example, you might choose to grant the role to your technical architect and a lead developer.</p>
+    <h3 class="heading-small">Org billing manager</h3>
+    <p>An Org billing manager is a point of contact for billing</p>
     <h3 class="heading-small">Org auditor</h3>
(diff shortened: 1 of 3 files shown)


src/users/permissions.njk
```diff
@@ -2,20 +2,20 @@
...
-      Org level role descriptions
+      What can these roles do?
     </span>
   </summary>
 
   <div class="govuk-details__text" aria-hidden="true">
-    <h3 class="heading-small">Org billing manager</h3>
-    <p>An Org billing manager is a point of contact for billing</p>
     <h3 class="heading-small">Org manager</h3>
     <p>An Org manager can create/delete spaces and edit user roles. The Org Managers would typically be senior staff: for example, you might choose to grant the role to your technical architect and a lead developer.</p>
```
paola-roccuzzo: Please change to
<p>An org manager can create and delete spaces, and edit user roles. Org managers are typically senior team members, like a technical architect or a lead developer.</p>

src/users/users.njk
```diff
@@ -63,10 +62,10 @@
   <thead class="govuk-table__head">
     <tr class="govuk-table__row">
       <th class="govuk-table__header name" scope="col">Name</th>
-      <th class="govuk-table__header is-billing-manager" scope="col">Org billing manager</th>
       <th class="govuk-table__header is-org-manager" scope="col">Org manager</th>
+      <th class="govuk-table__header is-billing-manager" scope="col">Org billing manager</th>
       <th class="govuk-table__header is-org-auditor" scope="col">Org auditor</th>
-      <th class="govuk-table__header spaces" scope="col">Spaces a collaborator on</th>
+      <th class="govuk-table__header spaces" scope="col">Spaces assigned or can access</th>
```
paola-roccuzzo: Spaces they can access

(as a minimum they can all access, some people might have further rights - 'Spaces assigned' means something else...)

PR #75, opened 2018-07-02
Restructure the application

paroxp:
What
----

At current, it may be really difficult to look around the source code
and come up with what's going on.

We're defining the difference between the components and libraries.

We're renaming `statement` to `statements`, so it matches the "standard"
we're having among the other components.

The `express-middleware` has been moved from `router` library to `app`
component, as it's really a bridge between the two and belongs in our
"custom" implementation as oppose to the library.

How to review
-------------

- Make sure the tests pass
- Make sure navigating around still brings what's expected
- See if the structure makes more sense - complain, comment and ask to re-juggle things


carolinegreen:
This is really nice and makes more sense to me.  I was forever looking for the express-middleware in app rather than the router library!

PR #77, opened 2018-07-13
Add admin fee

carolinegreen:
What
----

Add 10% fee to calculator so that people aren't surprised at the price.  Refactor some of the CSS so that it is easier to add styles.  Also capitalise "vat".

How to review
-------------

Run the application, try adding and removing things from the calculator and check the additional fee is correct.

Who can review
---------------

Anyone who isn't me.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -243,7 +243,7 @@
     "@types/qs": {
       "version": "6.5.1",
       "resolved": "https://registry.npmjs.org/@types/qs/-/qs-6.5.1.tgz",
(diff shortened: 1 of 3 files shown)


alext:
A style comment: the "Package.lock" commit should really have a commit message explaining why that change is being made.

chrisfarms:
There appears to be a fight going on with package-lock.json between different versions of npm... I'm not really sure how to resolve that... nvm doesn't take care of npm version just the node version by the looks of things

carolinegreen:
Yes, I was putting in just package lock as it is a change that happens between npm versions.  It isn't really changing any of the main dependencies, and isn't intended to really be looked at later on.  I don't know if we should at least pin the npm version in the package.json as currently it just says 5 and above, and I am using a newer version.

PR #78, opened 2018-07-13
Scrap HMR and just use nodemon when watching build

chrisfarms:
## What

HMR broke again 🙄 

We have been having constant issues with hot-module-reloading not
working or breaking whenever we make unrelated changes.

This gives up on the HMR approch and instead uses nodemon to watch the
dist dir and restart the server whenever it changes. This has a tiny
delay compared to HMR and will mean the server connection gets dropped momentarily but at least it actually works.

A nodemon webpack plugin is used to make the config simpler.

## How to review

* Run it locally
* Try changing a njk template or something
* watch it restart the server 

## Who can review

Not @chrisfarms


tlwr:
This works really well for me, but unless I am misunderstanding node `require` (which wouldn't be the first time), `nodemon-webpack-plugin` is always required, but it will only be installed in dev.

paroxp:
In `package.json` it is defined in the `devDependencies` section. It is also `require`d in the config file that is used by `webpack`, which is only to be run on your local machine or the build CI. In production it will not be loaded nor installed :)

tlwr:
Ok cool, thanks for explaining :)

PR #79, opened 2018-07-16
[#158974168] Cost reports aggregating multiple orgs

chrisfarms:
What
----

We have created a new report page (`/report/cost/YYYY-MM`) that gives an overview of the costs of all orgs you are authorised to see billing for.

How to review
-------------

Run the thing, run the tests, view a report.

Who can review
---------------

Anyone


tlwr:
screencapture-localhost-3000-reports-cost-2018-07-2018-07-16-13_20_05

src/components/reports/cost.ts
```diff
@@ -0,0 +1,169 @@
...
+    accessToken: ctx.token.accessToken,
+    apiEndpoint: ctx.app.cloudFoundryAPI,
+  });
+
+  const orgs = await cf.organizations();
+  const orgGUIDs = orgs.map(o => o.metadata.guid);
+
+  const orgQuotas: {[key: string]: IOrganizationQuota} = (await Promise.all(
+    orgs.map(async (org: IOrganization) => {
+      const quota = await cf.organizationQuota(org.entity.quota_definition_guid);
```
paroxp: Now, this _probably_ won't ever be a problem to our tenants, but it could be to admin users. I'm wondering if `Promise.all()` would be a better thing to do, meaning the calls are parallel as oppose to asynchronous.

- Not a blocker

src/main.ts
```diff
@@ -68,8 +68,12 @@ async function main() {
     port: parseInt(process.env.PORT || '0', 10),
   });
 
-  process.once('SIGINT', () => server.stop());
-  process.once('SIGTERM', () => server.stop());
+  process.once('SIGINT', () => {
+    server.stop().catch(err => console.error(err));
```
paroxp: May be able to replace this with:

```js
server.stop().catch(console.error);
```

- Not a blocker

src/server/server.test.ts
```diff
@@ -61,14 +61,14 @@ describe('server test suite', () => {
...
+    return server.stop();
   });
 
   it('should fail to start an already started server', async () => {
     const server = new Server(express());
     await server.start();
     await expect(server.start()).rejects.toThrow(/already started/);
-    server.stop();
+    return server.stop();
   });
```
paroxp: I knew I missed few places in the tests... Shoutout for finding and fixing these! 👏

src/components/reports/cost.ts
```diff
@@ -0,0 +1,169 @@
...
+      incVAT, exVAT,
+    };
+  });
+}
+
+export function createQuotaCostRecords(
+  records: ReadonlyArray<IOrgCostRecord>,
+): ReadonlyArray<IQuotaCostRecord> {
+  const costsByQuota = records
+  .reduce((
```
keymon: I wonder if it is worth it to iterate across all events and aggregate all the cost for non existent orgs and quotas

src/components/reports/cost.ts
```diff
@@ -0,0 +1,169 @@
...
+      }, {});
+}
+
+export function createOrgCostRecords(
+  orgs: ReadonlyArray<IOrganization>,
+  orgQuotas: {[key: string]: IOrganizationQuota},
+  orgBillableEvents: {[key: string]: ReadonlyArray<IBillableEvent>},
+): ReadonlyArray<IOrgCostRecord> {
+  return orgs
+  .map(org => {
```
keymon: I wonder if it is worth it to iterate across all events and aggregate all the cost for non existent orgs and quotas

src/components/reports/cost.ts
chrisfarms: Maybe I misunderstand, but we _are_ doing this in parallel.

The orgs.map is a collection of promises, and then we wait for them all to resolve.

paroxp: Yeah, it's right there in line 52... Never mind, ignore.

src/main.ts
chrisfarms: I got into a habit of avoiding passing references of `console.{log,error}` functions around as they sometimes error with "illegal invocation", I think it's an issue with how they generate the stack frames for the log entry (like new Error)

src/components/reports/cost.ts
chrisfarms: Interesting. Due to the way you have to request events for a specific org, I would hope that there would be no "left-over" events... and so a change to paas-billing would need to be made for that.

PR #80, opened 2018-07-18
Initialise GOV.UK scripts on paas-admin

paroxp:
What
----

We've faced an issue where the mobile verion of the site would not
function correctly - specifically when it comes to mobile menu - due to
JavaScript not being initialised.

How to review
-------------

- Run application
- Shrink the window to mobile view
- See if the menu trigger works


carolinegreen:
LGTM 👍

PR #81, opened 2018-07-31
[#156911499] Implement basic paas-admin acceptance tests

keymon:
https://www.pivotaltracker.com/story/show/156911499

What
----

We wan to add some integration tests for paas-admin which modify and make assertions against a real Cloud Foundry environment.

The current tests which use HTTP mocks don't accurately represent Cloud Controller's functionality. For example, they aren't able to catch the bug that prevented org managers from inviting users in #156761419. You need to do manual testing in a browser to be confident that there are no existing bugs or new regressions.

In this PR we implement the initial scaffolding for the integration test with some basic tests. We will add more tests in follow up stories 

See commits for details.

Dependencies
------------------

Once this merged, update and review https://github.com/alphagov/paas-cf/pull/1463

How to review
-------------

Code review. 

Test/Review together with https://github.com/alphagov/paas-cf/pull/1463

Who can review
---------------

Anyone

### DIFF
--- a/acceptance-tests/main.test.ts
+++ b/acceptance-tests/main.test.ts
@@ -1,0 +1,130 @@
+import {describe,it,before,after} from 'mocha';
+import expect from 'expect';
+import axios from 'axios';
+import CloudFoundryClient from '../src/lib/cf/cf';
+import UAAClient, { authenticateUser } from '../src/lib/uaa';
+import Browser from 'zombie';
+
+const {
+  PAAS_ADMIN_BASE_URL,
+  CF_API_BASE_URL,
+  ADMIN_USERNAME,
+  ADMIN_PASSWORD,
+} = process.env;
+if (!PAAS_ADMIN_BASE_URL) { throw 'PAAS_ADMIN_BASE_URL environment variable not set' }
+if (!ADMIN_USERNAME) { throw 'ADMIN_USERNAME environment variable not set' }
+if (!ADMIN_PASSWORD) { throw 'ADMIN_PASSWORD environment variable not set' }
+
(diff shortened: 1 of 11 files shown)


acceptance-tests/main.test.ts
```diff
@@ -0,0 +1,130 @@
+import {describe,it,before,after} from 'mocha';
```
paroxp: Was there any particular reason to install `mocha`? Is something missing from jest that is already installed and used?

acceptance-tests/main.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import expect from 'expect';
```
paroxp: Again, additional dependency that is already introduced by jest.

acceptance-tests/main.test.ts
keymon: Jest did not work OK with zombie. We commented it in the commit.

acceptance-tests/main.test.ts
keymon: We wanted to use the jest matchers, despite using mocha.

keymon:
> I may be on the wrong here, but you seem to have introduced new functionality that isn't really used in the application but the integration tests itself. Such creating and deleting orgs and users. I'm not sure if it's the right thing to do. Wouldn't you rather set these up and tear down outside of the application?

On one hand yes, you are right here. But on the other hand these methods are very likely useful within paas-admin or even if extract these clients.

richardTowers:
Sorry @paroxp - I didn't see your comments when I hit merge 🤦‍♂️

PR #82, opened 2018-07-31
[#156911499] Accept any pending documents

richardTowers:
What
----

We discovered that in some environments there are some documents that
the user should agree to. These break the tests because you can't use
paas-admin until you've agreed to the documents.

Adding some setup that uses the accounts API to accept the documents.

How to review
-------------

Just a code review should be fine

Who can review
---------------

Anyone technical


keymon:
LGTM

PR #84, opened 2018-08-07
Fix contrast issue

paroxp:
What
----

There is an issue, where all elements provided by the govuk-frontend will
have a yellow background set on :focus. It's particularly affecting links,
buttons and summary boxes. These elements also have a blue text in them,
which could cause difficulty to our users when it comes to reading the
content. It means we don't meet the required WCAG 2 AA standard [1].

Design system team thinks the roll out may take a long time, due to the
nature of it being used in the wide government and have no issues with
us overriding it with the provided black colour.

This is only a temporary solution and should be reverted back once the
issue upstream is resolved. [2]

[1]: https://snook.ca/technical/colour_contrast/colour.html#fg=005EA5,bg=FFBF47
[2]: https://github.com/alphagov/govuk-frontend/issues/944

How to review
-------------

- See if change makes sense
- Run the application and play around to see if nothing is negatively affected
- Run a standards check with the colours

### DIFF
--- a/src/layouts/govuk.screen.scss
+++ b/src/layouts/govuk.screen.scss
@@ -17,6 +17,8 @@
 
 @import '../components/statements/statements';
 
+@import './contrast-fix';
+
 .text-right {
   text-align: right;
 }
--- a/src/layouts/_contrast-fix.scss
+++ b/src/layouts/_contrast-fix.scss
@@ -1,0 +1,22 @@
+// There is an issue, where all elements provided by the govuk-frontend will
+// have a yellow background set on :focus. It's particularly affecting links,
+// buttons and summary boxes. These elements also have a blue text in them,
+// which could cause difficulty to our users when it comes to reading the
+// content. It means we don't meet the required WCAG 2 AA standard [1].
+
+// Design system team thinks the roll out may take a long time, due to the
+// nature of it being used in the wide government and have no issues with
+// us overriding it with the provided black colour.
+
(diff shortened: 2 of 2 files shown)


src/layouts/govuk.screen.scss
```diff
@@ -17,6 +17,8 @@
 
 @import '../components/statements/statements';
 
+@import './tmp';
```
carolinegreen: I think we should call this colour contrast so we don't forget what it is.  In my experience, temporary things often aren't very temporary.

paroxp: Fair.

paroxp:
Nice touch, love it. Id be vary of adding new colours to the palette though. Before I go ahead changing it, don't you think we should stick to "3 colour" rule? :D

Happy either way.

carolinegreen:
I think the three colour rule came from consolidating colours, and at one point there were more colours.  I think too, that patterns are meant to evolve and the current ones aren't meeting our needs.  If we call the colour variable something sensible and we have the explanation next to it, it should be okay.

paroxp:
Addressed.

I didn't make it a variable, as I hear naming things makes it harder to get rid of the thing. People get attached and stuff.

carolinegreen:
Haha, I think naming it would be better as otherwise we are more likely to end up with arbitrary colours.

paroxp:
Fine. Readability - I'm all up for it.

carolinegreen:
Thanks :).

PR #86, opened 2018-08-14
[#159649991]Add friendlier message for billing being unavailable

carolinegreen:
What
----

Change error message from 'Sorry an error occurred' to 'Billing is currently unavailable, please try again later' when billing is down.

How to review
-------------

Run the application, break the billing API, check that it errors as expected.

Who can review
---------------

Anyone who isn't me.

### DIFF
--- a/src/components/statements/statements.ts
+++ b/src/components/statements/statements.ts
@@ -12,7 +12,6 @@ import {
   CLOUD_CONTROLLER_READ_ONLY_ADMIN,
 } from '../auth';
 
-import errorsTemplate from '../errors/error.500.njk';
 import usageTemplate from './statements.njk';
 
 export const adminFee = .1;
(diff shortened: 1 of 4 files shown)


src/components/statements/statements.ts
```diff
@@ -88,7 +89,16 @@ export async function viewStatement(ctx: IContext, params: IParameters): Promise
...
+  let events;
+  try {
+    events = await billingClient.getBillableEvents(filter);
+  } catch {
+    return {
+      body: errorsTemplate.render({
+        errorMessage: 'Billing is currently unavailable, please try again later.',
+      }),
+    };
+  }
```
paroxp: I'm wondering if that's the best approach. This would require us to load the errors template in (not currently) all of the modules, and try and catch... Can we `throw new CustomErrorWhithAMessageThatItsDefinitionDoesntExistYet('Billing is currently unavailable, please try again later.')` and let the middleware handle it?

keymon:
Can we get the ID of the corresponding story in the title of the PR? as;

*[#123456789] Add friendlier message for billing being unavailable* 

This way it is easier to correlate to the stories during review and the updates of the PR also update the story in pivotal. 

If it is not related to any story, add something like *[#SUPPORT]* or *[#MAINTENANCE]* or so to tell that it is intentional.

carolinegreen:
@keymon Yes, I'm bad for forgetting about that.

keymon:
> @keymon Yes, I'm bad for forgetting about that.

Yep, you get used to add it at the end.

BTW: What is the story # in this case?

PR #89, opened 2018-08-23
[#157220369] Speed up data collection

paroxp:
What
----

We're parallelising a lot of different calls into a single promise to
prevent blocking each other.

We're also reducing the need for spaceSummary call where possible, as it
appears it's one of the slower transactions.

Reduces about 1/3 of time waiting.

In order to be consistent and not to pull all the CUPS at once, we'd
like to introduce a space filter if possible. The same signature is for
the services.

How to review
-------------

- Code review
- Run an application and have a poke around


src/components/spaces/_tabs.njk
```diff
@@ -5,7 +5,7 @@
       spaceGUID: space.metadata.guid
     }) }}">
       Applications
-      <span class="number-of-items">{{ space.entity.apps | length }}</span>
+      {# <span class="number-of-items">{{ space.entity.apps | length }}</span> #}
```
chrisfarms: Is this commenting out intentional?

chrisfarms: ok yeah decided to scrap the labels for now

PR #91, opened 2018-08-24
Upgrade govuk frontend

paroxp:
What
----

In order not to fall behind, we're upgrading the govuk-frontend package
by itself.

The way we do this, has now changed, therefore we're cleaning up the
README file.

We'd also like to upgrade the dependencies. These upgrades should 
cover any outstanding CVEs too.

After the dependancy upgrade, the linter started to point out some
issues. These are now solved with the use of `npm run fix:ts`

Finally, we're setting the `$govuk-global-styles` true for SCSS.

In case we'd forget using the `govuk-link` class, the GOV.UK Frontend
has us covered.

How to review
-------------

- Expect tests to pass
- Run the application
- Have a poke around and see if nothing stands out

---

In loving memory of #88


src/main.ts
```diff
@@ -68,6 +68,9 @@ async function main() {
     port: parseInt(process.env.PORT || '0', 10),
   });
 
+  process.on('uncaughtException', (err) => {
```
chrisfarms: This sets off my spidey-sense 🕸 

Why is this required?

src/server/server.test.ts
```diff
@@ -34,7 +34,7 @@ describe('server test suite', () => {
     const port = server.http.address().port;
     await request(`http://localhost:${port}`).get('/').expect(200);
     await server.stop();
-    expect(await request(`http://localhost:${port}`).get('/').timeout(500)).toBeUndefined();
+    return expect(request(`http://localhost:${port}`).get('/').timeout(500)).rejects.toThrow(/ECONNREFUSED/);
```
chrisfarms: does toThrow actually return a promise? this looks slightly odd but I'm a bit green with this library

src/main.ts
paroxp: Been so long. I don't think it is. I've put it there as I was debugging something and it actually helped out. Must have forgotten to take it out.

src/server/server.test.ts
paroxp: It would appear so. I had real good time trying to solve that test.

chrisfarms: ok cool

PR #92, opened 2018-08-24
[#160021128] Refactor the backing services listing

paroxp:
What
----

Now that we've made some recent changes to speed up admin tool, we've
forgotten that we require some extra fields for the backing service
listing.

These are now being added and the template is re-adjusted to fit the
purpose.

Test case has been added to prevent similar mistake happening again to
the same component.

How to review
-------------

- Run application
- Navigate to `admin` > `billing` > `Backing Services`
- Experience the page rendering correctly


keymon:
What I do not understand is why it was not logging any error

keymon:
Merging as the feature is broken in production and the code only touches this endpoint.

PR #99, opened 2018-10-05
[#160984488] Add CSRF protection

paroxp:
What
----

We'd like to make sure the application is well protected against the
[CSRF]. @richardTowers has well spotted that `helmet` middleware is not
doing that for us, meaning we need to take an action to assure ourselves
we deliver secure software.

The `csurf` npm package is a middleware that can do things for us
without much effort.

After introduction of `csurf` package and the implementation, the thests
started to fail due to missing function.

This is a hacky way to assure the function exists while not all of the
middlewares have been setup.

Yet again, this shows lacks in implementation of the tests.

[CSRF]: https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)

How to review
-------------

⚠️ ~~**I have not tested this, since I don't have my environment running.**~~ ⚠️

- Install dependencies
- Code review
- Run the application
- Navigate around as much as possible to make sure nothing is restricted
- Attempt to send few forms, particularly `user invite` and `permission management`
- Attempt to submit a form from external service (see code sample below for help)

Code sample
------------

Make sure to fill in the URL properly.

```html
<html>
  <body>
    <form action="__PLEASE_PROVIDE_URL_TO_INVITE_PAGE__" method="post">
      Email: <input type="email" name="email" value="evil@hacker.com">
      <br>
      <button type="submit">powah!</button>
    </form>
  </body>
</html>
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -237,6 +237,16 @@
       "integrity": "sha512-EIjmpvnHj+T4nMcKwHwxZKUfDmphIKJc2qnEMhSoOvr1lYEQpuRKRz8orWr//krYIIArS/KGGLfL2YGVUYXmIA==",
       "dev": true
     },
+    "@types/csurf": {
+      "version": "1.9.35",
+      "resolved": "https://registry.npmjs.org/@types/csurf/-/csurf-1.9.35.tgz",
+      "integrity": "sha512-2EVN+Bt2Vd8u+11xeJ64BjCYVOlhqaob82FPAw8VzOOWAYfP8TFvB7RD67CShEz45JXiI+38mlNJHKrArCzFMw==",
+      "dev": true,
+      "requires": {
+        "@types/express": "*",
+        "@types/express-serve-static-core": "*"
+      }
+    },
     "@types/events": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@types/events/-/events-1.2.0.tgz",
@@ -2336,6 +2346,16 @@
         "randomfill": "^1.0.3"
       }
     },
+    "csrf": {
+      "version": "3.0.6",
+      "resolved": "https://registry.npmjs.org/csrf/-/csrf-3.0.6.tgz",
+      "integrity": "sha1-thEg3c7q/JHnbtUxO7XAsmZ7cQo=",
+      "requires": {
+        "rndm": "1.2.0",
+        "tsscmp": "1.0.5",
+        "uid-safe": "2.1.4"
+      }
+    },
     "css-loader": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/css-loader/-/css-loader-1.0.0.tgz",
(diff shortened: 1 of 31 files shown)


keymon:
I do not think this is working, it is unable to verify the token. Maybe it is because the problem described in this comment: https://github.com/expressjs/csurf/issues/52#issuecomment-74426892 so that csrf must be done after parsing the bodt.

paroxp:
Cool.

Now have tested (thanks Alex) and can confirm it works as expected.

keymon:
It fails when submitting the aggrements.

To reproduce:
```
echo '{"content": "# Foo\n\nblablablaa\n"}' > agreement.json

DEPLOY_ENV=alext
ACCOUNTS_PASSWORD="$(aws s3 cp s3://gds-paas-${DEPLOY_ENV}-state/cf-vars-store.yml  - | awk '/secrets_paas_accounts_admin_password/ {print $2}')"
ACCOUNTS_URL=https://accounts.${DEPLOY_ENV}.dev.cloudpipeline.digital
curl -u admin:${ACCOUNTS_PASSWORD}     -H "Content-Type: application/json"     -X PUT -d @./agreement.json     ${ACCOUNTS_URL}/documents/TnC

``` 

Then visit the admin app and click on accept.

It fails with: `ForbiddenError: invalid csrf token`

keymon:
This makes that work: 

```diff
$ git diff 
diff --git a/src/components/terms/middleware.ts b/src/components/terms/middleware.ts
index 3362a81..59eee11 100644
--- a/src/components/terms/middleware.ts
+++ b/src/components/terms/middleware.ts
@@ -22,7 +22,10 @@ export function termsCheckerMiddleware(config: IAccountsClientConfig): express.H
 
   app.get('/agreements/:name', sync(async (req, res) => {
     const document = await accounts.getDocument(req.params.name);
-    res.send(termsTemplate.render({document}));
+    res.send(termsTemplate.render({
+      document: document,
+      csrf: req.csrfToken(),
+    }));
   }));
 
   app.post('/agreements', sync(async (req, res) => {
```

But maybe we should test it and add some sort of test to check that the terms can be accepted at least.

Also, maybe you want to refactor the terms part to include inside of the router.ts :thinking:

paroxp:
Good spot!

I honestly don't know why is the terms module so much separated from the rest of the app... In fact that's the reason I didn't initially spot that handler.

I think it's to big of a change for this PR.

keymon:
I am sorry being a PITA, but test don't pass... :man_shrugging:

keymon:
This can make the test pass:

```
$ git diff
diff --git a/src/components/terms/middleware.test.ts b/src/components/terms/middleware.test.ts
index be7158c..0a6ef97 100644
--- a/src/components/terms/middleware.test.ts
+++ b/src/components/terms/middleware.test.ts
@@ -26,6 +26,7 @@ app.use((req: any, _res: any, next: any) => {
     user_id: fakeUserID,
   }, tokenKeys[0]);
   req.token = new Token(accessToken, tokenKeys);
+  req.csrfToken = function csrfToken () { return "foo" }
   next();
 });
 
```

This would be the simplest, otherwise you need to refactor the test to get the csrf token in the post. 

But we are, again, ignoring and working around the feature in the tests instead of testing it.  Although this is what has been done in other tests.

paroxp:
The ideal approach is refactoring this module to our own router style... This would then be testing the `csurf` library along with it... It was a real nightmare to make it work with the library at current state, so I decided against it and did something similar to your suggestion...

We could be mocking the check if that makes you feel any better?

Integration tests

PR #105, opened 2018-10-12
Do not include CSRF for calculator

keymon:
What
----

The calculator is a GET method and does not require CSRF.

But including it might cause problems in the long run because
people share the calculator urls.

How to review
-------------

Code review

Who can review
---------------

Not me

### DIFF
--- a/src/components/calculator/calculator.njk
+++ b/src/components/calculator/calculator.njk
@@ -48,7 +48,6 @@
                 <td class="govuk-table__cell" scope="row">{{ serviceName }} {{ version }}</td>
                 <td class="govuk-table__cell ">
                   <form class="paas-service-selection" method="get">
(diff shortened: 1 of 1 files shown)


src/components/calculator/calculator.ts
```diff
@@ -101,7 +101,6 @@ export async function getCalculator(ctx: IContext, params: IParameters): Promise
     body: calculatorTemplate.render({
       routePartOf: ctx.routePartOf,
       linkTo: ctx.linkTo,
-      csrf: ctx.csrf,
```
paroxp: Can we not remove that line?

Though it's not being used, it's still parsed in the header over at: https://github.com/alphagov/paas-admin/blob/master/src/layouts/govuk.njk#L18
I would imagine on the long run it would be problematic to debug if the time comes.

keymon: sure

PR #106, opened 2018-10-18
[#161315584] Include all spaces / plans in filter dropdowns

richardTowers:
What
----

Previously we were filtering the items down to the selected space / plan
and then getting the list of spaces / plans from the filtered items.
Consequently the list of spaces / plans didn't include all the things it
should have.

A couple of nice extras

* we can switch from `let` to `const` for `items` because we're not
  reassigning it anymore.
* we can use `filter` instead of `reduce` and it's much simpler (and
  probably a bit more efficient)

How to review
-------------

* Code review
* Deploy to a dev environment (or locally) and test (steps to repro in
  [#161315584](https://www.pivotaltracker.com/story/show/161315584))

Who can review
---------------

Anyone

### DIFF
--- a/src/components/statements/statements.test.ts
+++ b/src/components/statements/statements.test.ts
@@ -104,6 +104,30 @@ describe('statements test suite', () => {
     expect(response.body).toContain('batman');
   });
 
+  it ('populates filter dropdowns with all spaces / services', async () => {
+    // tslint:disable:max-line-length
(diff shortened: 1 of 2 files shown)


LeePorte:
I'm not seeing the filter at all after deploying this in dev.

statement

LeePorte:
Operator error, disregard previous comment

PR #114, opened 2018-11-14
[#160508552] Add location banner

henrytk:
What
----

Add a location tag for every page on pazmin which lets people know which platform they are using (AWS region).

How to review
-------------

* Deploy:
  * locally change the paas-admin resource to pin this branch instead of a tag 
  * run `make dev pipelines`
  * run `make dev run_job JOB=post-deploy`
* manually check every feasible page, because we don't have proper tests. Good luck.
* I haven't tested the Terms page yet (`/agreements/:name`).
  **Note:** I tried creating a user in prod and wasn't forced to make the agreement, so I think we may have already broken this feature.

I don't like the way I have implemented this. I am worried any page would break silently (no tests would fail) if the way the parameter gets passed in changes. The user experience for such an error would be to see a blank page. Not good. Ideally I want some kind of default filter (`location | defaul('unknown')`) to at least stop it breaking (the CSS selector for `-unknown` could be `display:none;`). Such a filter is used elsewhere in this repository, but I couldn't get it to work for me. I am not much of a NodeJS person.

Who can review
---------------

Anybody, but I would like @richardTowers opinion on it.

### DIFF
--- a/src/components/errors/error.ts
+++ b/src/components/errors/error.ts
@@ -7,6 +7,18 @@ import internalServerError from './error.500.njk';
 
 export class UserFriendlyError extends Error {}
 
+/* istanbul ignore next */
+function platformLocation(region: string): string {
+  switch (region) {
+    case 'eu-west-1':
+      return 'Ireland';
+    case 'eu-west-2':
+      return 'London';
+    default:
+      return region;
+  }
+}
+
 export function internalServerErrorMiddleware(err: Error, req: any, res: express.Response, next: express.NextFunction) {
   req.log.error(err);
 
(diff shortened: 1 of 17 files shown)


src/components/errors/error.ts
```diff
@@ -7,6 +7,18 @@ import internalServerError from './error.500.njk';
 
 export class UserFriendlyError extends Error {}
 
+/* istanbul ignore next */
+function platformLocation(region: string): string {
```
keymon: I think we can get rid of this. There are two options:

The `internalServerErrorMiddleware` is setup in the app.ts, where we know already the location. It is also called from some controllers with the location.

The `pageNotFoundMiddleware` is also setup in the controller.

I suggest we create some constructor for these two middlewares, where we pass the `location` to be used.

Maybe we can put the `internalServerErrorMiddleware` middleware in the `app.ctx` so that can be used from the controllers ([auth](https://github.com/alphagov/paas-admin/blob/294cb8b068c72f1153ce19958e71698d158cacaf/src/components/auth/auth.ts#L25-L26) and [terms](https://github.com/alphagov/paas-admin/blob/294cb8b068c72f1153ce19958e71698d158cacaf/src/components/terms/middleware.ts#L15-L16))

richardTowers:
Not sure whether it's a good idea or not, but you could avoid inlining the HTML by doing the tag in pure CSS:

```
{% block header %}
  {{ govukHeader({
    homepageUrl: "/",
    containerClasses: "govuk-width-container paas-region-{{region}}",
    ...
```

```
.paas-region-london .govuk-header__content::after {
    content: 'London';
    float: right;
    position: absolute;
    top: 0 /*some value*/;
    right: 0 /*some value*/;
}
```

```
.paas-region-ireland .govuk-header__content::after {
    content: 'Ireland';
    float: right;
    position: absolute;
    top: 0 /*some value*/;
    right: 0 /*some value*/;
}
```

henrytk:
@richardTowers I gave your suggestion a try. The following:

```
{% block header %}
  {{ govukHeader({
    homepageUrl: "/",
    containerClasses: "govuk-width-container paas-region-{{region}}",
    ...
```

will result in the literal `-{{region}}` being passed to the function.

henrytk:
@richardTowers I tried your suggestion and it seems to render the `-{{region}}` literally.

PR #116, opened 2018-12-19
[#162443556] Change method of obtaining user spaces

tnwhitwell:
What
----

Rather than relying on Cloud Foundry's API to provide all the spaces a user has permissions for, this determines them programmatically.

1 Get all spaces in the org
2 Get all users for those spaces
3 Build an object, keyed by user, with roles and spaces

Also, updated the template as the object's structure has changed.

How to review
-------------
1. Code Review
2. Deploy and test:
   1. Deploy to a dev environment, and route `admin.` to it
   2. Log in as an Organisation manager
   3. Confirm you can see Spaces for each user in `/organisations/org_guid/users`

Who can review
---------------

NOT @tnwhitwell, @richardTowers


src/components/users/users.ts
```diff
@@ -163,31 +179,39 @@ export async function listUsers(ctx: IContext, params: IParameters): Promise<IRe
...
+    userOrgRoles.reduce((acc, user) => ({...acc, [user.metadata.guid]: user}), {});
 
-    return userWithSpaces;
+  const spaceUserLists = await Promise.all(spacesVisibleToUser.map(async space => {
+    return {space, users: await cf.usersForSpace(space.metadata.guid)};
   }));
 
+  const userRolesByGuid: IUserRolesByGuid = {};
+
+  for (const spaceUsers of spaceUserLists) {
```
richardTowers: I've been thinking about it a bit more, and I think we should pull this big complicated bit out into a function which can be tested independently.

One bug I can think of at the moment is: if a user is in the org, but isn't in any spaces that the logged in user can see, they won't be able to see their org roles (because they're only assigned inside the spaces loop).

tnwhitwell: > I've been thinking about it a bit more, and I think we should pull this big complicated bit out into a function which can be tested independently.

Agreed. I'll take a look at that

> if a user is in the org, but isn't in any spaces that the logged in user can see, they won't be able to see their org roles

Actually, it seems worse than that. I'll move this PR back to WIP

src/components/users/users.njk
```diff
@@ -71,36 +71,35 @@
...
   <tbody class="govuk-table__body">
-    {% for user in users %}
+    {% for guid, user in users %}
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">
         {% if isAdmin or isManager %}
-          <a href="{{ linkTo('admin.organizations.users.edit', {organizationGUID: organization.metadata.guid, userGUID: user.metadata.guid}) }}" class="govuk-link">
-            {{ user.entity.username }}
+          <a href="{{ linkTo('admin.organizations.users.edit', {organizationGUID: organization.metadata.guid, userGUID: guid}) }}" class="govuk-link">
+            {{ user.username }}
```
paroxp: I'm not sure what's the reasoning for shortening/flattening the data is, but we've decided back in the day to do it the long way `user.[entity,metadata].property` in order to keep the consistency with the API as it confused people a little.

I don't feel strongly really, prefer the shorten way. But it's still worth to keep in mind.

src/lib/cf/cf.test.data.ts
```diff
@@ -823,8 +823,8 @@ export const userRolesForSpace = `{
   "resources": [
     {
       "metadata": {
-        "guid": "uaa-id-363",
-        "url": "/v2/users/uaa-id-363",
+        "guid": "uaa-id-253",
```
paroxp: I am really sorry you had to deal with this...

This, is possibly the worst thing about pazmin... We were hoping to introduce `jest` and `mock` which would not require the data to exist... This didn't happen due to prioritisation...

If you decide to keep this app alive and have capacity, I would suggest doing that sooner rather than later for a more pleasant experience...

src/components/users/users.njk
tnwhitwell: It was really just to reduce the provided data a little: rather than shoving around big chunky objects, just the elements we know we need are sent to the template.

Believe it's ok, but good to keep in mind.

src/components/users/users.ts
```diff
@@ -151,6 +172,27 @@ async function setAllUserRolesForOrg(
   ));
 }
 
+export function _getUserRolesByGuid(
+    userOrgRoles: ReadonlyArray<IOrganizationUserRoles>,
+    spaceUserLists: ReadonlyArray<ISpaceUsers>,
+  ): IUserRolesByGuid {
+
+  const spacesByUser =
```
bandesz: during testing it came out, that this map contains maximum one space for any user

richardTowers: I've addressed this now. This case is covered by unit tests now, and I've tested it in my dev environment.

src/components/users/users.njk
```diff
@@ -71,36 +71,35 @@
...
+        {% if 'org_auditor' in user.orgRoles %}
           <img class="tick" alt="Yes" src="./tick.png">
         {% endif %}
       </td>
       <td class="govuk-table__cell">
-        {% for space in user.spaces %}
-          <a href="{{ linkTo('admin.organizations.spaces.applications.list', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid}) }}" class="govuk-link">{{ space.entity.name }}</a>
-        {% endfor %}
+        {% set comma = joiner(", ") %}
+        {% for space in user.spaces %}{{ comma() }}<a href="{{ linkTo('admin.organizations.spaces.applications.list', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid}) }}" class="govuk-link">{{ space.entity.name }}</a>{% endfor %}
```
AP-Hunt: I think this line is too long to be readable. I believe you could split it over multiple lines and have the same output, although I'm not so familiar with this template language.

```
{% for space in user.spaces %}
{{ comma() }}
<a href="{{ linkTo('admin.organizations.spaces.applications.list', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid}) }}" 
    class="govuk-link">
{{ space.entity.name }}
</a>
{% endfor %}
```

src/components/users/users.ts
```diff
@@ -163,31 +204,21 @@ export async function listUsers(ctx: IContext, params: IParameters): Promise<IRe
...
     cf.organization(params.organizationGUID),
     cf.usersForOrganization(params.organizationGUID),
   ]);
 
-  const usersWithSpaces = await Promise.all(users.map(async (user: IOrganizationUserRoles) => {
-    const userWithSpaces = {
-      ...user,
-      spaces: new Array(),
-    };
+  const spacesVisibleToUser = await cf.spaces(params.organizationGUID);
```
AP-Hunt: Since this doesn't rely on anything returned in the promises in L207, we could group L214 in with it. As it is, we're performing an extra await for no extra value. Unless there's a reason this has to be separated?

richardTowers: Yep, you're absolutely right - this is just sequential for no reason. I'll update it.

PR #120, opened 2019-02-19
[#163084536] Removal of an org permission from a user instead removes all org permissions

tnwhitwell:
What
----

When any single org permission is removed from a user (ie. the checkbox is unchecked), all org permissions were removed from that user.

When a checkbox is unchecked in a HTML form, the `value` is just not serialised and sent, rather than a 0 being sent.

Instead of checking for truthiness of the new value, we now just check to see if the POSTed form value is undefined.

How to review
-------------

* Code review
* Build paasmin from this branch, and deploy to an environment. Add permissions to a user, then remove them - you should see it behaving as expected (permission is added and removed accordingly), vs previous behaviour (permission is removed, along with all others)

Who can review
---------------

Anyone but me and @mogds

### DIFF
--- a/src/components/users/users.test.ts
+++ b/src/components/users/users.test.ts
@@ -373,7 +373,7 @@ describe('users test suite', async () => {
     expect(response.body).toContain('Confirm user deletion');
   });
 
-  it('should update the user, set BillingManager role and show success - User Edit', async () => {
+  it('should update the user, set BillingManager role and show success - User Edit', async () => { // TODO: fix label
     const response = await users.deleteUser(ctx, {
       organizationGUID: '3deb9f04-b449-4f94-b3dd-c73cefe5b275',
       userGUID: '5ff19d4c-8fa0-4d74-94e0-52eac86d55a8',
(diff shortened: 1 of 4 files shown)


src/components/users/users.test.ts
```diff
@@ -429,6 +429,32 @@ describe('users test suite', async () => {
     expect(response.body).toContain('Updated a team member');
   });
 
+  it('should update the user, remove BillingManager role and show success - User Edit', async () => {
+    const scope = nock(ctx.app.cloudFoundryAPI)
+    // tslint:disable:max-line-length
+    .delete('/v2/organizations/3deb9f04-b449-4f94-b3dd-c73cefe5b275/billing_managers/uaa-user-changeperms-123456?recursive=true').reply(200, `{}`)
```
AP-Hunt: These URIs are pretty hefty, and scattered through this file, with repeated GUIDs.

We could reduce the mental and maintenance burden by extracting the repeated GUIDs in to variables and using [template strings](https://www.typescriptlang.org/docs/handbook/basic-types.html#string) to cleanly build the URIs.

src/components/users/users.test.ts
```diff
@@ -429,6 +429,32 @@ describe('users test suite', async () => {
...
+      // tslint:enable:max-line-length
+    ;
+    const response = await users.updateUser(ctx, {
+      organizationGUID: '3deb9f04-b449-4f94-b3dd-c73cefe5b275',
+      userGUID: 'uaa-user-changeperms-123456',
+    }, {
+      org_roles: {
+        '3deb9f04-b449-4f94-b3dd-c73cefe5b275': composeOrgRoles({
```
AP-Hunt: Same here. Extract the GUIDs?

AP-Hunt:
Code looks OK, and the bug is fixed. Merging.

PR #124, opened 2019-04-12
Fix test 'should store a session in a signed cookie'

AP-Hunt:
What
----
This test has always had the correct expectation, but of the wrong cookie.

How to review
-------------
1. Code review
Travis tests should take care of the rest.

Who can review
---------------
Anyone

### DIFF
--- a/src/components/app/app.test.ts
+++ b/src/components/app/app.test.ts
@@ -189,7 +189,9 @@ describe('app test suite', () => {
     it('should store a session in a signed cookie', async () => {
       const response = await request(app).get('/test');
 
-      expect(response.header['set-cookie'][0]).toContain('pazmin-session.sig');
+      expect(response.header['set-cookie'].some(
+        (x: string) => x.includes('pazmin-session.sig')))
+        .toBe(true);
     });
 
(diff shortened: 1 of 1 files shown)


src/components/app/app.test.ts
```diff
@@ -189,7 +189,7 @@ describe('app test suite', () => {
     it('should store a session in a signed cookie', async () => {
       const response = await request(app).get('/test');
 
-      expect(response.header['set-cookie'][0]).toContain('pazmin-session.sig');
```
richardTowers: Travis is 💢about this. Maybe `expect(response.header['set-cookie'].some(x => x.include('pazmin-session.sig'))).toBe(true)`

src/components/app/app.test.ts
```diff
@@ -189,7 +189,9 @@ describe('app test suite', () => {
...
+      expect(response.header['set-cookie'].some(
+        (x: string) => x.indexOf('pazmin-session.sig') > -1))
```
richardTowers: This is fine, but you could use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

PR #125, opened 2019-04-12
[#164763735] Pull quota from the billing response

mogds:
What
----

The paas-admin cost report (`/reports/cost/2018-01-01`) currently queries the cf api for the quota guid and the billing endpoint for the events. Therefore the quota doesn't reflect the quota applied during the billing period.

Paas-billing has been [updated](https://github.com/alphagov/paas-billing/pull/64) to include the quota_definition_guid so it has the correct quota. This updates the cost report to pull the quota from the billing response if available.

Merge Steps
-------------

**Deploy https://github.com/alphagov/paas-billing/pull/64 before merging this PR**

How to review
-------------

- Code review 
- Run tests

Who can review
---------------

Not me, @AP-Hunt


src/lib/billing/types.ts
```diff
@@ -83,6 +83,7 @@ interface IUsageEvent {
 
 interface IBillableEvent extends IUsageEvent {
   readonly price: IPrice;
+  readonly quotaGUID: string;
```
richardTowers: This might be better as a nullable type (I can't remember if the syntax is `string?` or `string | null`) because not all billable events will have quotas on them (presumably, since we only just started writing them).

src/components/reports/cost.test.ts
```diff
@@ -88,6 +88,7 @@ describe('cost report test suite', () => {
         "org_guid": "7f9c0e11-e7f1-41d7-9d3f-cb9d05110f9e",
         "space_guid": "2e030634-2640-4535-88ed-e67235b52ceb",
         "plan_guid": "f4d4b95a-f55e-4593-8d54-3364c25798c4",
+        "quota_definition_guid":"3f2dd80c-7dfb-4e7f-b8a9-406b0b8abfa3",
```
AP-Hunt: Formatting

src/components/reports/cost.test.ts
```diff
@@ -153,6 +154,7 @@ describe('cost report test suite', () => {
         "org_guid": "a7aff246-5f5b-4cf8-87d8-f316053e4a20",
         "space_guid": "2e030634-2640-4535-88ed-e67235b52ceb",
         "plan_guid": "f4d4b95a-f55e-4593-8d54-3364c25798c4",
+        "quota_definition_guid":"dcb680a9-b190-4838-a3d2-b84aa17517a6",
```
AP-Hunt: Formatting

src/components/reports/cost.test.ts
```diff
@@ -206,7 +208,7 @@ describe('cost report test suite', () => {
       {
         eventGUID: '', eventStart: new Date(), eventStop: new Date(),
         resourceGUID: '', resourceName: '', resourceType: '', orgGUID: '',
-        spaceGUID: '', spaceName: '', planGUID: '', numberOfNodes: 0, memoryInMB: 0,
+        spaceGUID: '', quotaGUID: '', spaceName: '', planGUID: '', numberOfNodes: 0, memoryInMB: 0,
```
AP-Hunt: Could we pull creating these objects out in to a helper function that mixes-in its input? That should help tidy up these tests a bit, and make them easier to maintain in the long run.

```
function newBillableEvent(overrides){
    return mixin({billable: event}, overrides);
}

const summed = reports.sumRecords([
    newBillableEvent({price: {incVat: 10}}),
    newBillableEvent({price: {incVat: 5.5}}),
]);
```

src/lib/billing/billing.test.ts
```diff
@@ -20,6 +20,7 @@ describe('lib/billing test suite', () => {
         "org_guid": "7f9c0e11-e7f1-41d7-9d3f-cb9d05110f9e",
         "space_guid": "2e030634-2640-4535-88ed-e67235b52ceb",
         "plan_guid": "f4d4b95a-f55e-4593-8d54-3364c25798c4",
+        "quota_definition_guid":"3f2dd80c-7dfb-4e7f-b8a9-406b0b8abfa3",
```
AP-Hunt: Formatting

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "strict": true,
+    "strict": false,
```
AP-Hunt: Out of curiosity, why are we introducing the changes to the `strict` and `noImplictAny` transpiler flag changes?

src/components/reports/cost.test.ts
richardTowers: There's nice syntax for this in the language even:

```
const defaultBillableEvent = { eventGuid: '', /*... and the rest */ }

const summed = reports.sumRecords([
  {...defaultBillableEvent, price: {incVat: 10}},
  {...defaultBillableEvent, price: {incVat: 5.5}},
])
```

(or something like that)

tsconfig.json
mogds: Set that due to npm build errrors. I've changed it back but have added `strictNullChecks` and set this to false. 

Similarly, set `noImplicitAny` to false due to `Object literal's property 'details' implicitly has an 'any[]' type.` errors.

PR #127, opened 2019-04-17
[#165345614] Add S3 Buckets to the pricing calculator

46bit:
What
----

Includes S3 buckets in the pricing calculator. See commit message for full details.

How to review
-------------

* Code review;
* See if you are happy with how it is displayed in my environment **TODO: LINK**.

Who can review
---------------

Not @46bit.

### DIFF
--- a/src/components/calculator/calculator.njk
+++ b/src/components/calculator/calculator.njk
@@ -34,7 +34,7 @@
     </div>
   </div>
 
-  <div class ="govuk-grid-row">
+  <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds paas-service-list">
       <table class="govuk-table">
         <thead>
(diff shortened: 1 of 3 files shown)


46bit:
`aws-s3-bucket default` won't work as the name of the row in the Pricing Calculator. It isn't descriptive, but more importantly its length breaks the `app` row onto two lines:

Screen Shot 2019-04-17 at 14 54 16

46bit:
I think this is a *bit* better:

Screen Shot 2019-04-17 at 16 22 15

46bit:
There are no tests that it prints `default` if there are also non-default plans. That is because there is complex logic for how the plaintext pricing plan names are transformed, and such a test would be a nightmare to put together.

tlwr:
Screen Shot 2019-04-18 at 17 18 02

4️⃣:two::zero:

PR #136, opened 2019-04-17
Remediate OOM killing on PaaS by increasing memory

tlwr:
What
----

Scale up PaaSmin with more memory, to stop it from being OOM-killed when people access the bills for large orgs

How to review
-------------

Code review

Who can review
---------------

Not @tlwr


tlwr:
🤠

PR #148, opened 2019-04-18
Request logging

tlwr:
What
----

- Stops prettifying logs as this hides some useful output
- Adds request/response interceptors for axios
- Adds the interceptors to every HTTP client except UAA
- Abuses the type system (sorry)

How to review
-------------

- Code review
- Spin up paas-admin (`LOG_LEVEL=info` PLEASE) and check the logs (very verbose now)

Who can review
---------------

Not @tlwr


src/lib/axios-logger/axios.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function newAxiosErrorInterceptor(name: string, logger: BaseLogger) {
+  return (err: any) => {
+    logger.error({message: `${name} encountered error`});
+
+    return err;
+  };
+}
+
+export function Intercept(inst: AxiosInstance, name: string, log: BaseLogger) {
```
richardTowers: The convention in JS is that functions only start with capital letters if they're constructors. This looks like a bit of the Golang style guide has sneaked in :)

richardTowers:
How does it abuse the type system? 🤔

tlwr:
It abuses the type system with a malicious `any` type to avoid making a new class wrapping Axios

https://github.com/alphagov/paas-admin/pull/148/files#diff-457a1abb40b4f2d0f22f8fdb78f27a43R13

which is bad but consistent with other places in the codebase

src/lib/axios-logger/axios.ts
tlwr: You are not wrong

tlwr:
I'm going to raise the comments here in a separate PR

richardTowers:
> It abuses the type system with a malicious `any` type to avoid making a new class wrapping Axios
> 
> https://github.com/alphagov/paas-admin/pull/148/files#diff-457a1abb40b4f2d0f22f8fdb78f27a43R13
> 
> which is bad but consistent with other places in the codebase

Ah yeah, that's fine :)

The compiler is our servant, not our master.

PR #159, opened 2019-04-23
Refactor Calculator's use of pricing plan names

46bit:
What
----

See commit message.

How to review
-------------

Who can review
---------------

### DIFF
--- a/src/lib/billing/types.ts
+++ b/src/lib/billing/types.ts
@@ -22,14 +22,70 @@ interface IPricingPlanResponse {
 }
 
 interface IPricingPlan {
-  readonly planName: string;
+  readonly name: string;
+  readonly metadata: IPricingPlanMetadata;
   readonly planGUID: string;
-  readonly serviceName: string;
   readonly validFrom: Date;
   readonly components: ReadonlyArray<IComponent>;
   readonly numberOfNodes: number;
   readonly memoryInMB: number;
   readonly storageInMB: number;
+}
+
+// Our Billing API provides one unstructured name field for each pricing plan.
+// We parse that into structured metadata about the plan. Here's an example.
+//
+// 1. The Billing API gives us this pricing plan name:
+//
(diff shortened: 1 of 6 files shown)


src/lib/billing/types.ts
```diff
@@ -22,16 +22,72 @@ interface IPricingPlanResponse {
...
+//     planVersion = "default"
+//     planVariant = ""
+//
+// By clustering pricing plans by serviceName and planVersion, we can then render
+// a pretty Pricing Calculator.
+//
+// FIXME: extracting metadata from the planName is not a reliable
+// method, and we do not want to have to give paas-admin scopes to look up
+// this information in cf. A better solution is to get paas-billing to
+// provide this information.
```
richardTowers: Massive 👍 for this huge comment ❤️

46bit: I guess that's why you gave me Senior ❤️😄

46bit: I am tempted to drop most of the changes in this PR and add that comment's content to the README. I think the best fix is to add this metadata to `paas-billing` itself, and that a story to do that would be better than finishing this PR.

richardTowers: Agree, it would be nice to solve this closer to the source - paas-billing is abdicating a lot of responsibility by just spitting out basically raw config as an "API".

That said, this code might be harder to write in Go, so perhaps this PR is still the pragmatic approach for something which isn't a major issue.

tlwr:
This looks good

We haven't imported the types (or at least the Travis log alleges this), let me know if you want to delegate

46bit:
Thanks @tlwr. I'm doing lots of work on `paas-billing` and will try to move this metadata into that so that this PR isn't needed.

46bit:
Closing in favour of https://github.com/alphagov/paas-cf/pull/2243, which solves the underlying problem 🎉

PR #160, opened 2019-04-23
Use javascript naming for request log interceptor

tlwr:
What
----

s/Intercept/intercept

How to review
-------------

Code review

Git grep

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/lib/accounts/accounts.ts
+++ b/src/lib/accounts/accounts.ts
@@ -2,7 +2,7 @@ import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
 import moment from 'moment';
 import { BaseLogger } from 'pino';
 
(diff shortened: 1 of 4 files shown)


src/lib/accounts/accounts.ts
```diff
@@ -2,7 +2,7 @@ import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
 import moment from 'moment';
 import { BaseLogger } from 'pino';
 
-import {Intercept} from '../axios-logger/axios';
+import {intercept} from '../axios-logger/axios';
```
AP-Hunt: The imported function is imported on `src/lib/axios-logger/axios.ts#L49` as `Intercept`

PR #171, opened 2019-04-25
bau: add APM-server performance and error monitoring

tlwr:
What
----

Adds the ability to use an Elastic APM server to collect performance metrics and exceptions/traces.

How to review
-------------

- Code review
- Run APM server locally (or use https://govuk-paas-apm-server.london.cloudapps.digital) and run paas-admin locally with APM enabled

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,4 +1,40 @@
 import pino from 'pino';
+const logger = pino({
+  level: process.env.LOG_LEVEL || 'info',
+});
+
+import apm from 'elastic-apm-node';
+/* istanbul ignore next */
+{
+  const isApmEnabled = ['yes', 'true'].includes((process.env.ENABLE_APM || '').toLowerCase());
+
+  const apmServiceName = process.env.APM_SERVICE_NAME || 'paas-admin-local';
+
+  const apmSecretToken = process.env.APM_SECRET_TOKEN;
+  if (isApmEnabled && !apmSecretToken) {
+    throw new Error('APM_SECRET_TOKEN is a required environment variable');
+  }
+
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -1,4 +1,39 @@
 import pino from 'pino';
+const logger = pino({
+  level: process.env.LOG_LEVEL || 'info',
+});
+
+import apm from 'elastic-apm-node';
+/* istanbul ignore next */
+{
+  const isApmEnabled = !!process.env.ENABLE_APM;
```
richardTowers: Could be a bit surprising that `ENABLE_APM=false ...` leads to `!!'false' === true`

tlwr: Changed, thoughts?

tlwr:
I'm marking this as not draft, I think we can get it on master and deployed without doing anything more

This will allow us to connect it to the live APM server temporarily without making changes to the pipeline.

src/main.ts
```diff
@@ -1,4 +1,40 @@
...
+    throw new Error('APM_SERVER_URL is a required environment variable');
+  }
+
+  if (isApmEnabled) {
+    logger.info({
+      message: `Starting apm with name ${apmServiceName} -> ${apmServerUrl}`,
+    });
+  }
+
+  apm.start({
```
richardTowers: Should this be in the `if (isApmEnabled)` block? or will it just do nothing if it's called with an empty URL

tlwr: There is an `active` key in the object passed to start, so other bits of the code where we talk to `apm` don't have to be wrapped in an if statement

PR #173, opened 2019-04-25
Support: Fix bug showing allocations

tlwr:
What
----

### Fix bug in org summary

The org summary space shows a summary for each space:

Screen Shot 2019-04-25 at 14 45 09
*the GOV.UK Notify org summary page*

The above number is the total amount of memory used for each app assuming that the number of instances for each app is one.

Now the calculation includes the number of instances.

### Add additional columns in space summary

The space summary page shows the amount of memory and disk used for each app:

Screen Shot 2019-04-25 at 14 47 21
*the GOV.UK Notify production space summary

Now, if the number of instances is not 1, the total amount is calculated:

Screen Shot 2019-04-25 at 15 41 34
*what it looks like with stub-api*

How to review
-------------

- Code review
- Spin up with stub-api locally (see README) and check

Who can review
---------------

Pair with @46bit

### DIFF
--- a/src/components/spaces/applications.njk
+++ b/src/components/spaces/applications.njk
@@ -37,13 +37,19 @@
             <th class="govuk-table__header" scope="row">
               <a href="{{ linkTo('admin.organizations.spaces.applications.view', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, applicationGUID: application.metadata.guid}) }}" class="govuk-link">{{ application.entity.name }}</a>
             </th>
-            <td class="govuk-table__cell ">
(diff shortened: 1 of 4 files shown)


46bit:
I don't think the brackets convey enough meaning to make clear what the bigger number is 🙁

tlwr:
I tried a few other options:

- additional columns (breaks when routes are too long)
- `1gb (Total: 2gb)` looked a bit ugly

Thoughts on just making total a `title=` so it appears on mouseover and screenreaders?

tlwr:
@46bit and I 🍐d on this IRL

PR #211, opened 2019-05-03
Show stack on app pages

46bit:
What
----

Adds the Stack name to the application page.

How to review
-------------

Code review.

Who can review
---------------

Not @46bit


46bit:
Working in my dev environment:

### An app known to be using `cflinuxfs3`

Screen Shot 2019-05-03 at 16 45 45

### An app known to be using `cflinuxfs2`

Screen Shot 2019-05-03 at 16 47 08

tlwr:
For docker apps this will show `Stack: cflinuxfs3` (or whatever is default) even though they technically don't have a stack.

This is okay in my opinion but we may want to hide the stack and/or show the docker image. Although this could be in a separate PR.

Other than that looks good to me

46bit:
> For docker apps this will show Stack: cflinuxfs3 (or whatever is default) even though they technically don't have a stack.
>
> This is okay in my opinion but we may want to hide the stack and/or show the docker image. Although this could be in a separate PR.

That is a good point. I'm not going to fix that here but happy to review such a PR.

PR #254, opened 2019-05-23
[#166027792] Implement SSO uplift journey for GDS users

AP-Hunt:
What
----
Google SSO is currently only available to GDS users. This commit implements a user journey to allow eligible users to enable SSO for their account for themselves, rather than by contacting support.

How to review
-------------
Code review

When your environment is running...
1. Checkout this branch in `paas-admin`
2. `cf t -o admin -s public`
3. `npm run push` to push over the existing app in your space.

When the app is deployed...
1. Create a new user in CF with the uaa origin (default)
2. Sign in to paas-admin as that user
3. Run through the journey.

Who can review
---------------
Not @AP-Hunt 
Journey reviewed by Debs Durojaiye

### DIFF
--- a/src/components/account/account.test.ts
+++ b/src/components/account/account.test.ts
@@ -1,13 +1,12 @@
 import jwt from 'jsonwebtoken';
 import nock, {RequestBodyMatcher} from 'nock';
-import uaa, {IUaaUser} from '../../lib/uaa';
 
+import pino = require('pino');
 import * as uaaData from '../../lib/uaa/uaa.test.data';
 import {IContext} from '../app';
 import {config} from '../app/app.test.config';
 import {Token} from '../auth';
 import * as account from './account';
-import pino = require('pino');
 
 describe('account test suite', () => {
   let ctx: IContext;
@@ -33,7 +32,8 @@ describe('account test suite', () => {
 
       it('should contain an explanation of the process, and a link to the docs', async () => {
         const response = await account.getUseGoogleSSO(ctx, {});
-        expect(response.body).toContain('https://docs.cloud.service.gov.uk/get_started.html#use-the-single-sign-on-function');
+        expect(response.body)
+          .toContain('https://docs.cloud.service.gov.uk/get_started.html#use-the-single-sign-on-function');
         expect(response.body).toContain('id="sso-process-explanation"');
       });
     });
@@ -47,7 +47,7 @@ describe('account test suite', () => {
         const response = await account.getUseGoogleSSO(ctx, {});
         expect(response.body).toContain('id="opt-out-process-explanation"');
       });
-    })
+    });
   });
 
   describe('account.use-google-sso.post', () => {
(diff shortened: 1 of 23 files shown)


src/components/organizations/organizations.ts
```diff
@@ -21,13 +21,16 @@ export async function listOrganizations(ctx: IContext, _params: IParameters): Pr
 
   const organizations = await cf.organizations().then(sortOrganizationsByName);
 
+  const user = await account.fetchLoggedInUser(ctx);
```
richardTowers: We could do:

```
const organizations, user = await Promise.all(
  cf.organizations().then(sortOrganizationsByName),
  account.fetchLoggedInUser(ctx),
);
```

And save ourselves a bit of time making the requests sequentially.

src/lib/billing/billing.ts
```diff
@@ -194,7 +194,7 @@ function parseBillableEvent(ev: IBillableEventResponse): IBillableEvent {
 }
 
 function parsePricingPlan(plan: IPricingPlanResponse): IPricingPlan {
-  // FIXME: extracting the serviceName from the planName is not a reliable
+  // FIX_ME: extracting the serviceName from the planName is not a reliable
```
richardTowers: `FIXME` / `FIX_ME`, potato potato

src/components/account/account.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    describe('name', () => {
+      it('name combines user\'s given and family names', () => {
+        // We don't have a full implementation of this type
+        // because it's used to represent a JSON payload
+        const uaaUser = {} as any as IUaaUser;
+        uaaUser.name = {
+          givenName: 'User',
+          familyName: 'Name',
+        };
```
richardTowers: Don't really like `as any as IUaaUser`. I'd would do:

```
const uaaUser: any = {
  name: {
    givenName: 'User',
    familyName: 'Name',
  }
};
```

... and leave it as the `any` type. It's only a test, and downcasting it back to `IUaaUser` doesn't really buy you any type safety.

src/components/account/account.test.ts
```diff
@@ -0,0 +1,182 @@
...
+    describe('authenticationMethod', () => {
+      it('returns u&p if the user\'s origin is uaa', () => {
+        const uaaUser = {} as any as IUaaUser;
+        uaaUser.origin = 'uaa';
+
+        const acctUser = new account.AccountUser(uaaUser);
+        expect(acctUser.authenticationMethod).toEqual('Username & password');
+      });
+
+      it('returns google if the user\'s origin is uaa', () => {
```
richardTowers:
```suggestion
      it('returns google if the user\'s origin is google', () => {
```

src/lib/billing/billing.ts
richardTowers: Ah, I see what you're doing - you're silencing the linter warnings by using a pattern it doesn't recognise as suspicious. Trés sneaky.

richardTowers: Going to amend it to `NOTE` - same effect, but less sneaky.

PR #265, opened 2019-05-29
Allow the filtering of cost reports by service

richardTowers:
What
----

This allows you to go to `/reports/cost/2019-03-01?service=blah` and see
only the bills for that particular service.

The particular use case for this is to see how much we've billed for
elasticsearch.

It works in a pretty hacky way, using the query param as a substring to
search through all the plan names.

How to review
-------------

* Code review
* Run with stub api and visit `localhost:3000/reports/cost/2019-03-01?service=wizarding-plan` (or one of the other magical plans) and confirm that the bills change as expected

Who can review
---------------

Not @richardTowers


AP-Hunt:
All working correctly locally

PR #271, opened 2019-06-03
Fix colour of the ireland tag

richardTowers:
What
----

When we made the region tags into links we broke the specificity, so
Ireland is currently showing as blue. It clearly should be green since
Ireland is the Emerald Isle.

How to review
-------------

* Code review

Who can review
---------------

Not rich


46bit:
😦 oh dear, my moving fast mostly broke things

PR #284, opened 2019-06-17
[#166335176] Use paas-accounts for username data

AP-Hunt:
What
----
Two major things in this PR

## Thing the first
Expanding which endpoints we cover in our stubbing of the different APIs means we've run in to conflicts, where two distinct APIs have the same path component and different responses. This commit spins up the different stub APIs on different ports so that those conflicts don't happen.

It also extracts setting the required environment variables in to a script, so as not to pollute `package.json`.

## Thing the second
Querying `paas-accounts` for a user's username, and falling back to the username stored in UAA.

Usernames are only shown in 2 places in `paas-admin` right now, and other than that what's introduced here, we only use `paas-accounts` for agreements-related actions. For that reason, I've added the fetching of the username from `paas-accounts` to the 2 places it's needed. 

If we begin doing more with `paas-accounts`, I think we should move it to be part of the library code under `src/lib/`

How to review
-------------
**But wait!** If https://github.com/alphagov/paas-cf/pull/1952 hasn't been merged when you come to review this, you'll need to deploy it first. Or use the stub APIs to test. Probably a good idea to do both.

1. Code review `8852fccb1774a1e60d3a8e1c8483e53d6e620e9f`
2. Code review `51203b803dba1a4a82912ffe276c51653b1cdea1` and `2138fd6493f2fecdb1626a5b3122193b4c129b47`
3. Deploy `paas-admin` from this branch and test it.

Who can review
---------------

### DIFF
--- a/src/components/spaces/spaces.ts
+++ b/src/components/spaces/spaces.ts
@@ -1,15 +1,10 @@
+import {AccountsClient} from '../../lib/accounts';
 import CloudFoundryClient from '../../lib/cf';
-import {
-  IApplication,
-  IOrganizationUserRoles,
-  IRoute,
-  IServiceInstance,
-  ISpace,
-} from '../../lib/cf/types';
-import { IParameters, IResponse } from '../../lib/router';
+import {IApplication, IOrganizationUserRoles, IRoute, IServiceInstance, ISpace} from '../../lib/cf/types';
+import {IParameters, IResponse} from '../../lib/router';
 
-import { IContext } from '../app/context';
-import { CLOUD_CONTROLLER_ADMIN, CLOUD_CONTROLLER_GLOBAL_AUDITOR, CLOUD_CONTROLLER_READ_ONLY_ADMIN } from '../auth';
+import {IContext} from '../app/context';
+import {CLOUD_CONTROLLER_ADMIN, CLOUD_CONTROLLER_GLOBAL_AUDITOR, CLOUD_CONTROLLER_READ_ONLY_ADMIN} from '../auth';
 
 import spaceApplicationsTemplate from './applications.njk';
 import spaceBackingServicesTemplate from './backing-services.njk';
@@ -135,6 +130,12 @@ export async function listSpaces(ctx: IContext, params: IParameters): Promise<IR
     logger: ctx.app.logger,
   });
 
+  const accountsClient = new AccountsClient({
+    apiEndpoint: ctx.app.accountsAPI,
+    secret: ctx.app.accountsSecret,
+    logger: ctx.app.logger,
+  });
+
   const isAdmin = ctx.token.hasAnyScope(
     CLOUD_CONTROLLER_ADMIN,
     CLOUD_CONTROLLER_READ_ONLY_ADMIN,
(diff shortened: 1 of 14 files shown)


stub-api/index.ts
```diff
@@ -4,23 +4,60 @@ import stubAccounts from './stub-accounts';
...
+    app: null,
+  },
+];
 
-stubUaa(app, config);
-stubAccounts(app, config);
-stubBilling(app, config);
-stubCf(app, config);
+for (const api of apis) {
+  api.app = express();
```
richardTowers: Could `api.app` actually just be a local variable inside this loop? Not sure why it needs to be on `IStubServerConfig`.

stub-api/stub-accounts.ts
```diff
@@ -10,6 +10,31 @@ function mockAccounts(app: express.Application, _config: IStubServerPorts): expr
...
   });
 
+  // All users should have a paas-accounts entry.
+  // We need to gather the IDs from different places
+  // in our stub data, in order for everything to
+  // fit together.
+  const userIds = [];
+  userIds.push(uaaStubData.userId);
+  const cfUsers = JSON.parse(cfStubData.users);
+  for (const k in cfUsers.resources) {
```
richardTowers: I think this loop could just be:

```
const userIds = cfUsers.resources.map(x => x.metadata.guid)
```

src/components/spaces/spaces.ts
```diff
@@ -211,3 +214,33 @@ export async function listSpaces(ctx: IContext, params: IParameters): Promise<IR
...
   };
 }
+
+async function hydrateAccountsUsernames(
+  userRoles: ReadonlyArray<IOrganizationUserRoles>,
+  accountsClient: AccountsClient,
+): Promise<ReadonlyArray<IOrganizationUserRoles>> {
+
+  const users = [];
+  for (const user of userRoles) {
```
richardTowers: Could be done with `.map()` instead of pushing in a loop.

richardTowers:
Manual testing: looks like it works against my dev environment - I can still see names for users that have them in paas-accounts, and for users that don't. I can see in the logs that the API requests are being made. Also works with the stub API (good job updating that to solve the collision).

I'm vaguely concerned this won't perform very well for organisations with large numbers of users (because it needs to make one API call per user, and then doesn't do any caching), but happy to :shipit: and see.

Presumably this is blocked on alphagov/paas-cf#1952?

AP-Hunt:
Force pushed with changes to address code review comments

> I'm vaguely concerned this won't perform very well for organisations with large numbers of users (because it needs to make one API call per user, and then doesn't do any caching)

I considered that when working on it. I came to the conclusion that we'd need to add a wider change to support a caching mechanism that would persist at least in a user's session. I don't think a single user is ever looked up in paas-accounts >1 time on any page.

richardTowers:
Yeah, or we could add an API endpoint in paas-accounts to request multiple users at the same time... But I think let's ship it and see how the performance is so we can prioritise whether to bother.

AP-Hunt:
> Yeah, or we could add an API endpoint in paas-accounts to request multiple users at the same time... But I think let's ship it and see how the performance is so we can prioritise whether to bother.

Do you mean like this one https://github.com/alphagov/paas-accounts/blob/master/api/get_users_handler.go#L22 ?

richardTowers:
Let's ship it anyway, and look at using the API with better performance later.

PR #285, opened 2019-06-19
[#166028126] MS SSO flow

AP-Hunt:
What
----

This a very large commit, which touches a number of things. The three major
changes are:

  1. New environment variables for Microsoft Active Directory
     credentials, and application domain name
  2. Implementation of handling an Open ID Connect authorization code grant handshake
  3. UI flow for guiding users through the uplift.

For GDS users of the Google SSO provider, we were able to safely assume that
the username we held for them would be the same as their Google account email
address. However, for Microsoft users we know this isn't the case.

For the Microsoft SSO provder, we map the `oid` claim (which contains
Microsoft's identifier for the user) to our `userName` field in UAA. This means
that when we change a user's origin from `uaa` to `microsoft`, we
simultaneously need to change their username to match the value of the claim.
Otherwise, UAA will be unable to line up the two fields and identify the user,
and thus the user cannot log in.

In order to get the `oid` claim for a user, we need to have them log in with
Microsoft, and then request some information about the identity of the person.
For that, we use Open ID Connect flow to request an identity token. We
subequently pull the `oid` claim from the token, and update the `userName` and
`origin` in UAA in one API call.

In the future, we might choose to implement this OIDC flow for non-GDS Google
users. In that event, we will need to modify this flow and add some
configurable claim mapping.

How to review
-------------
1. Deploy this [`paas-cf` PR](https://github.com/alphagov/paas-cf/pull/1961) which contains this change, and the changes required in `paas-uaa-customized-boshrelease` and `paas-cf`
2. In your dev env, invite a new user
3. Log in as the new user
4. Follow the MS SSO uplift flow 
5. Log out
6. Log in with MS SSO, and see you're logged in to the same user account as before

Who can review
---------------
Anyone but @AP-Hunt and @mogds


src/components/account/account.ts
```diff
@@ -72,6 +48,135 @@ export async function postUseGoogleSSO(ctx: IContext, _params: IParameters): Pro
...
       location: ctx.app.location,
+      providerName: 'Google',
+    }),
+  };
+}
+
+export async function getUseMicrosoftSSO(ctx: IContext, _params: IParameters): Promise<IResponse> {
+  const cfgProvided = ctx.app.oidcProviders.find((p: IOIDCConfig) => {
+    return p.providerName === 'microsoft';
+  });
```
richardTowers: A bit more syntax than strictly necessary here - could do:

```js
  const cfgProvided = ctx.app.oidcProviders.find(p => p.providerName === 'microsoft');
```

src/components/account/account.ts
```diff
@@ -72,6 +48,135 @@ export async function postUseGoogleSSO(ctx: IContext, _params: IParameters): Pro
...
+      location: ctx.app.location,
+      user,
+    }),
+  };
+}
+
+export async function postUseMicrosoftSSO(ctx: IContext, _params: IParameters): Promise<IResponse> {
+  const cfgProvided = ctx.app.oidcProviders.find((p: IOIDCConfig) => {
+    return p.providerName === 'microsoft';
+  });
```
richardTowers:
```js
  const cfgProvided = ctx.app.oidcProviders.find(p => p.providerName === 'microsoft');
```

src/components/account/account.ts
```diff
@@ -72,6 +48,135 @@ export async function postUseGoogleSSO(ctx: IContext, _params: IParameters): Pro
...
+    apiEndpoint: ctx.app.uaaAPI,
+    clientCredentials: {
+      clientID: ctx.app.oauthClientID,
+      clientSecret: ctx.app.oauthClientSecret,
+    },
+  });
+
+  const cfgProvided = ctx.app.oidcProviders.find((p: IOIDCConfig) => {
+    return p.providerName === 'microsoft';
+  });
```
richardTowers: Again, could be a bit shorter. I wonder whether it would be worth making `oidcProviders` a `Map<'google'|'microsoft', IOIDCConfig>` instead of an array that we have to keep searching? 🤔

richardTowers: Maybe `Map<string, IOIDCConfig>` if you don't want the type system to pretend it's not possible to configure other providers, although as it stands there would be no point in configuring anything other than microsoft because the code always goes fishing by providerName.

src/components/app/context.ts
```diff
@@ -23,7 +23,9 @@ export function initContext(req: any, router: Router, route: Route, config: IApp
   return {
     app: config,
     routePartOf: (name: string) => route.definition.name === name || route.definition.name.startsWith(name),
-    linkTo: (name: string, params: IParameters = {}) => router.findByName(name).composeURL(params),
```
richardTowers: I think it would be better to put an extra `absoluteLinkTo(name, params)` function on the context, rather than adding the absolute parameter to `linkTo` - it's not obvious at the call site what the boolean flag does in `linkTo('blah.blah', undefined, true)`, whereas `absoluteLinkTo('blah.blah')` would be clear.

richardTowers:
Code review: looks good - just needs manual testing then I'll approve.

git style comment - I'm not a big fan of `Addresses` as the subject line - `git log --oneline` isn't very helpful with word commit subjects. Could we amend these commits to have a brief summary of what they do in the message (good to keep the links to the comments too though).

richardTowers:
(Discussed with Mo who explained the plan was to squash the commits addressing the review comments, which is fine - I've squashed them now since I'm happy with the code)

richardTowers:
Tested manually in my dev env using https://github.com/alphagov/paas-cf/pull/1961

This is approved now, but we shouldn't merge it until after that paas-cf change goes out (to add the UAA configuration and code changes).

PR #306, opened 2019-07-05
Add report page showing cost breakdown by service

richardTowers:
What
----

It's often useful to know how much individual organisations are spending
on groups of services (e.g. postgres or elasticsearch). This information
makes it easier to plan capacity and prioritise work on features that
are getting a lot of usage.

At the moment it's pretty painful to see this information in one place,
so this commit adds a new report page which will show the cost breakdown
by organisation and service for a given month.

With the stub API it looks something like this:

image

This new report should have roughly the same performance as the existing
`/reports/cost/` endpoint - possibly a bit faster as it doesn't need to
get the quotas.

How to review
-------------

* Code review
* Check the tests / linter pass in travis
* `npm run build` and then `cf push` to a dev environment
* Look at the new cost report page (e.g. https://admin.towers.dev.cloudpipeline.digital/reports/cost-by-service/2019-06-01)

Who can review
---------------

Not @richardtowers


tlwr:
Code review and run locally :shipit:

PR #307, opened 2019-07-07
Add a Sankey diagram report showing which orgs use which services

richardTowers:
What
----

**NOTE**: Follows on from #306 - will be easier to review once that's merged.

Add a billing report which contains a visualisation of which organisations are spending what on which services. With the stub API it looks like this:

image

How to review
-------------

* Review #306 first
* Code review
* Test with the stub API
* `npm run build` and `cf push` to a dev environment

Who can review
---------------

Not @richardTowers


tlwr:
Reviewed directly after #306

LGTM, my review of the sankey code was to look at the d3-sankey tutorial -.- but it seems straightforward given my slightly rusty d3.

I'm glad you changed the data returned by the stub because in my review of #306 I got `unknown` because of the `some-org-guid` entrant, but thought that changing the stub was unrelated to the PR.

:shipit:

PR #76, opened 2018-07-06
Introduce jest

paroxp:
What
----

We're refactoring the tests to use `jest` instead of types.

It brings many benefits, among which we:
- Could get rid of nock <3
- Can focus tests programatically
- Run tests faster
- Make them prettier
- Make them easier to understand for Ginkgo/rspec comers

**This PR also upgrades the govuk-frontend!**

We were running out of date, but also we could get rid of the legacy `govuk_template_jinja` module. This also means we can remove some of the hack-arounds we've provided for `webpack`.

How to review
-------------

- Code review
- Run application
- Navigate through and see if nothing is out of place
- Run tests

PR #324, opened 2019-07-12
Remove blue theme color

richardTowers:
What
----

Back in https://github.com/alphagov/paas-admin/pull/76/files#diff-a6f85c1191fe205b2ba09f3405e041a6R8 Rafal roguishly snuck a themeColor of `blue` into the template. The themeColor is used by chrome mobile to set the colour of the URL bar, but [`blue` is a gross color](https://html-color.codes/hex/0000ff).

We should remove the themeColor and let chrome do whatever the default is.

How to review
-------------

* Code review
* Look how gross this theme colour is!

image

Who can review
---------------

Not @richardTowers

### DIFF
--- a/src/layouts/govuk.njk
+++ b/src/layouts/govuk.njk
@@ -5,7 +5,6 @@
 {% extends "govuk-frontend/template.njk" %}
 
 {% set htmlLang = "en" %}
-{% set themeColor = "blue" %}
 {% set bodyClasses = "govuk-body" %}
 
 {% block pageTitle %}PaaS Administration Tool{% endblock %}


src/layouts/govuk.njk
```diff
@@ -5,7 +5,6 @@
 {% extends "govuk-frontend/template.njk" %}
 
 {% set htmlLang = "en" %}
-{% set themeColor = "blue" %}
```
46bit: `black` might be good, as that's the colour of the GOV.UK header

richardTowers: It's actually `#0b0c0c`, but yeah, let's go with that.

richardTowers: Conversation with Luke:

```
Richard Towers  [24 minutes ago]
You mean sets the colour to govuk-black?


Richard Towers  [24 minutes ago]
I’m not sure whether we should remove the theme colour entirely or set it black like GOV.UK does :thinking2:

Richard Towers  [23 minutes ago]
`blue` is definitely out though :blue-cap-crew:

luke.malcher  [19 minutes ago]
default is the best imo. ...
```

Suggests just removing it, so I've switched back to that.

richardTowers: (Although `#0b0c0c` is the default in govuk-frontend, so it will still be black)

PR #342, opened 2019-07-19
Put Origin in GA

tlwr:
This PR is branched off #337 - review that first

What
----

Put the authenticated user's `origin` from UAA in a Google analytics dimension, if it exists.

How to review
-------------

Code review

[Read](https://developers.google.com/analytics/devguides/collection/gtagjs/custom-dims-mets) on how `gtag` deals with custom dimensions

See Concourse

Who can review
---------------

@richardTowers

### DIFF
--- a/src/components/app/app.test.ts
+++ b/src/components/app/app.test.ts
@@ -173,7 +173,7 @@ describe('app test suite', () => {
           `default-src 'none'; style-src 'self' 'unsafe-inline'; ` +
           `script-src 'self' 'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=' ` +
           `'sha256-G29/qSW/JHHANtFhlrZVDZW1HOkCDRc78ggbqwwIJ2g=' www.google-analytics.com ` +
-          `www.googletagmanager.com 'sha256-3PKQVkjJP08zzrQZEDAfQN6ScCpPnpo3b3VEocvDsdg='; ` +
+          `www.googletagmanager.com 'sha256-JcBlICm6OEcFiWPFdHSsGDody7xi9jgOetfKR1FhO7Y='; ` +
           `img-src 'self' www.google-analytics.com; ` +
           `connect-src 'self' www.google-analytics.com; frame-src 'self'; font-src 'self' data:; ` +
           `object-src 'self'; media-src 'self'`,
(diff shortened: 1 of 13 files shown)


tlwr:
Rebased so the diff is more obvious

src/components/app/app.test.ts
```diff
@@ -208,6 +209,14 @@ describe('app test suite', () => {
 
       expect(response.redirect).toEqual(orgs.definition.path);
     });
+
+    it('should add a meta tag for origin', async () => {
+      const orgs = router.findByName('admin.organizations');
+      const response = await agent.get(orgs.definition.path);
+
+      expect(response.status).toEqual(200);
+      expect(response.text).toMatch('<meta name="origin" content="uaa" />');
```
richardTowers: I think `origin` might not be sufficiently specific for a meta tag (e.g. one might confuse it with "origin" as in "same origin policy") maybe `x-user-identity-origin` or something?

tlwr: Agreed, resolved in 75d7c63f323ae29cae6759f09afbd1004edf659d

src/layouts/govuk.njk
```diff
@@ -14,17 +14,26 @@
...
-  </script>
+    var dimensions = {};
 
-  <meta name="csrf-token" content="{{context.csrf.toString()}}" />
+    var originTag = document.querySelector('meta[name="x-user-identity-origin"]')
+    if (originTag && originTag.content) {
+      dimensions['dimension1'] = originTag.content;
+    }
+
+    gtag('config', 'UA-43115970-5', dimensions);
```
richardTowers: I'm not sure this is right according to the google docs [you linked](https://developers.google.com/analytics/devguides/collection/gtagjs/custom-dims-mets):

```js
// Configures custom dimension<Index> to use the custom parameter
// 'dimension_name' for 'GA_MEASUREMENT_ID', where <Index> is a number
// representing the index of the custom dimension.
gtag('config', 'GA_MEASUREMENT_ID', {
  'custom_map': {'dimension<Index>': 'dimension_name'}
});

// Sends the custom dimension to Google Analytics.
gtag('event', 'any_event_name', {'dimension_name': dimension_value});
```

Shouldn't it be:

```js
gtag('config', 'UA-43115970-5', {
  'custom_map': {'dimension1': 'origin'}
});
gtag('event', 'origin_dimension', {'origin': originTag.content});
```

Or something? I'm not 100% sure - the docs are quite confusing.

tlwr: After reviewing the page I linked I realised I linked a not-helpful page.

`custom_map` is renaming things in the console, along with the implicit event.

My understanding is calling `gtag` with `'config'` as the first argument does two things:

- configure `gtag` with the ID (`UA-*`) and with any additional properties
- implicitly send a pageview event

So calling `'config'` and then calling `'event'` AFAICT is two events, but we only want one.

I think what we want is:

```
gtag('set', 'dimension1', originTag.content);
gtag('config', 'UA-43115970-5')
```

which is longhand for

```
gtag('config', 'UA-43115970-5', dimensions);
```

---

Info about `'set'`, `'event'`, and `'config'` [here](https://developers.google.com/gtagjs/reference/api)

Gohugo.io uses Google analytics in this fashion, search for `dimension1` within `view-source:https://gohugo.io/`

PR #350, opened 2019-07-24
[#167511176] Fix incorrect numberOfNodes for HA services

richardTowers:
What
----

Some services (at the moment just redis medium-ha-3.2) have multiple
nodes, and the price depends on the number of nodes.

Currently the calculator serializes the "state" to a set of hidden
inputs, which are submitted along with the form when the user adds an
app or service.

Unfortunately, the `numberOfNodes` submitted is always
`plans[0].numberOfNodes` - i.e. the number of nodes for the smallest
plan, which is invariably 1.

This is tricky to fix with the current approach because when the user is
choosing a redis plan they only interact with one select, where they
pick the plan guid.

We can't completely avoid serializing numberOfNodes and memoryInMB into
the DOM, because these _can_ be set by the user for apps, and we need to
store the users selections in the state.

We can avoid serializing them for services though, because the correct
value will always still be available on the plan.

This commit continues to use the serialized numberOfNodes and
memoryInMB for apps, but stops serializing them for services and uses
the value from the plans instead.

As a side benefit, this will reduce the number of query parameters in
the URL, which will mean the calculator can work for larger numbers of
services.

How to review
-------------

* Code review
* Push paas-admin to a dev env and test the calculator manually (I've done this in my dev env - see for yourself that the redis costs twice what it does in prod: https://admin.towers.dev.cloudpipeline.digital/calculator?rangeStart=2019-07-01&rangeStop=2019-07-31&items%5B0%5D%5BplanGUID%5D=b6949ea7-5c98-4c69-b981-4b5318ea8b7c&items%5B0%5D%5BnumberOfNodes%5D=&items%5B0%5D%5BmemoryInMB%5D=&items%5B1%5D%5BplanGUID%5D=c84d1bef-9500-4ce9-88b2-c0bd421bbf8a)

Who can review
---------------

Not @richardtowers

### DIFF
--- a/src/components/calculator/calculator.ts
+++ b/src/components/calculator/calculator.ts
@@ -122,26 +122,44 @@ export async function getCalculator(ctx: IContext, params: IParameters): Promise
 async function getQuote(billing: BillingClient, state: ICalculatorState): Promise<IQuote> {
   const rangeStart = moment(state.rangeStart);
   const rangeStop = moment(state.rangeStop);
-  const usageEvents = state.items.reduce((events: IUsageEvent[], item: IResourceItem) => {
-    return [
-      ...events,
-      {
-        eventGUID: uuid.v1(),
-        resourceGUID: uuid.v4(),
-        resourceName: (state.plans.find(p => p.planGUID === item.planGUID) || {planName: 'unknown'}).planName,
-        resourceType: (state.plans.find(p => p.planGUID === item.planGUID) || {serviceName: 'unknown'}).serviceName,
-        orgGUID: '00000001-0000-0000-0000-000000000000',
-        spaceGUID: '00000001-0001-0000-0000-000000000000',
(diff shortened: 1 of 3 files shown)


src/components/calculator/calculator.ts
```diff
@@ -123,24 +123,42 @@ async function getQuote(billing: BillingClient, state: ICalculatorState): Promis
   const rangeStart = moment(state.rangeStart);
   const rangeStop = moment(state.rangeStop);
   const usageEvents = state.items.reduce((events: IUsageEvent[], item: IResourceItem) => {
```
46bit: You told me to tell you to try changing this from a `reduce` to a `map`.

richardTowers: Yeah, that is better. I've force pushed that update.

PR #358, opened 2019-08-01
Add organisations report

tlwr:
Context
----

We have [a gross script in paas-cf](https://github.com/alphagov/paas-cf/blob/master/scripts/org-usage-report.rb) for working out which organisations are expiring or abandoned, the business process is as follows:

- On the first of each month, we are told by IFTTT to run the script
- The support engineer clones the previous Pivotal Tracker story
- The support engineer runs the script in London and in Ireland
- The outputs of the script are copied into Pivotal Tracker
- The engagement team and the product manager review the output usually via a CSV

This is pointless bureaucracy / "busy work", the actual user need is to be effectively able to explore organisations currently on the platform.

This PR adds a new report which adds a new report, for organisation usage across the platform. It does not replicate the output of the script exactly, but other parts of PaaSmin can provide the same functionality.

The report is available on `/reports/organisations` and shows all orgs which aren't test generated and segregates them approximately according to billability.

⬇️ Shows a screenshot

Screen Shot 2019-08-01 at 19 46 03
_An image of the new organisations report_

How to review
-------------

Code review

`cf push` to your dev environment and do some clickops

[tlwr preview](https://admin.tlwr.dev.cloudpipeline.digital/reports/organisations)

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/components/reports/organizations.ts
+++ b/src/components/reports/organizations.ts
@@ -1,0 +1,126 @@
+import lodash from 'lodash';
+import moment from 'moment';
+
+import CloudFoundryClient from '../../lib/cf';
+import { IOrganization, IOrganizationQuota } from '../../lib/cf/types';
+import { IParameters, IResponse } from '../../lib/router';
+import { IContext } from '../app/context';
+
+import organizationsTemplate from './organizations.njk';
+
+function trialExpiryDate(creation: Date): Date {
+  return moment(creation).add(90, 'days').toDate();
+}
+
+function filterRealOrgs(
+  orgs: ReadonlyArray<IOrganization>,
+): ReadonlyArray<IOrganization> {
+  return orgs
+  .filter(org => {
+    const name = org.entity.name;
+
+    if (name.match(/^(CATS|ACC|BACC|SMOKE)-/)) {
+      return false;
+    }
+
+    if (name === 'admin') {
+      return false;
+    }
+
+    return true;
+  })
+  ;
+}
+
+function filterTrialOrgs(
+  trialQuotaGUID: string,
+  orgs: ReadonlyArray<IOrganization>,
+): ReadonlyArray<IOrganization> {
+  const trialOrgs = orgs.filter(
+    o => o.entity.quota_definition_guid === trialQuotaGUID,
+  );
+
+  // return the oldest orgs first
+  return lodash
+    .sortBy(trialOrgs, o => new Date(o.metadata.created_at))
+  ;
+}
+
+function filterBillableOrgs(
+  trialQuotaGUID: string,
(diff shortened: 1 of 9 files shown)


src/components/reports/organizations.ts
```diff
@@ -0,0 +1,126 @@
...
+  const trialOrgs = filterTrialOrgs(trialQuotaGUID, organizations);
+  const billableOrgs = filterBillableOrgs(trialQuotaGUID, organizations);
+
+  const orgQuotaMapping = orgQuotas.reduce((
+    mapping: {[key: string]: IOrganizationQuota},
+    orgQuota: IOrganizationQuota,
+  ) => {
+    mapping[orgQuota.metadata.guid] = orgQuota;
+    return mapping;
+  }, {});
```
richardTowers: Since you're using lodash, this is simpler as:

```js
const orgQuotaMapping = lodash.groupBy(orgQuotas, orgQuota => orgQuota.metadata.guid);
```

src/components/reports/organizations.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const orgTrialExpirys = trialOrgs.reduce((
+    mapping: {[key: string]: Date}, org: IOrganization,
+  ) => {
+    mapping[org.metadata.guid] = trialExpiryDate(
+      new Date(org.metadata.created_at),
+    );
+
+    return mapping;
+  }, {});
```
richardTowers: Pretty sure this is worse, but while we're lodashing, you could do:

```js
const orgTrialExpirys = lodash.mapValues(
  lodash.groupBy(trialOrgs, org => org.metadata.guid),
  orgs => orgs.map(org => trialExpiryDate(
    new Date(org.metadata.created_at),
  ))
)
```

src/components/reports/organizations.ts
tlwr: We can't use `groupBy` because the value type is not enumerable but a single value.

We can use `keyBy`, which I have done:

```
const orgQuotaMapping: {[key: string]: IOrganizationQuota} = lodash
  .keyBy(orgQuotas, q => q.metadata.guid)
;
```

src/components/reports/organizations.ts
tlwr: As above: we can't use `groupBy` because the value type is not enumerable but a single value.

We can use `keyBy`, which I have done:

```
const orgTrialExpirys: {[key: string]: Date} = lodash
  .chain(trialOrgs)
  .keyBy(org => org.metadata.guid)
  .mapValues(org => trialExpiryDate(new Date(org.metadata.created_at)))
  .value()
;
```

PR #364, opened 2019-08-05
Add origin to Org / Users / Team members

tlwr:
What
----

As an org manager or org auditor of an organisation I want to be able to
see who in my organisation has enabled single sign on, so I can enforce
a security policy.

I chose to introduce `ul.plain` because the regular `govuk-list` class
looks very strange on this page, when we move to GOV.UK Frontend 3.0 we
can address this page a bit more. I applied this class to the list of spaces because it was getting very messy for organisations with many spaces.

image
_Image of the updated Org / Users / Team Members page_

How to review
-------------

Code review

Concourse

[Preview](https://admin.tlwr.dev.cloudpipeline.digital/organisations/0fc56d0f-f83b-406c-bdca-953917911ba0/users)

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/components/org-users/org-users.njk
+++ b/src/components/org-users/org-users.njk
@@ -63,7 +63,12 @@
 <table class="govuk-table user-list">
   <thead class="govuk-table__head">
     <tr class="govuk-table__row">
-      <th class="govuk-table__header name" scope="col">Name</th>
+      <th class="govuk-table__header name" scope="col">Email address</th>
+      <th class="govuk-table__header authentication" scope="col">
+        {% if isAdmin or isManager %}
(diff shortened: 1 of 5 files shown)


src/components/org-users/org-users.njk
```diff
@@ -99,13 +113,16 @@
       </td>
       <td class="govuk-table__cell">
         {% set comma = joiner(", ") %}
```
richardTowers: Do we still need this line, now we're not using `comma()` on 103?

tlwr: :done:

PR #395, opened 2019-08-21
App test builder

andy-paine:
What
----

Remove the big JSON blobs for the CF `/v2/apps/:guid` API endpoints in tests/stub-api.
Replace instead with a fluent builder so that tests are more explicit about what the API should return and assertions are based on local constants rather than strings living in a file full o' big ol' JSON strings.
I plan to do the others like this too (app summary is now fairly low hanging fruit) but smaller commits == betterer

How to review
-------------

Let your CI system run the tests and assure yourself behaviour hasn't changed.
Run pazmin with the stub-api and check the spaces/application page still looks alright.

Who can review
---------------

Anyone on the PaaS team. Especially people who hate scrolling through JSON in Typescript strings.

### DIFF
--- a/src/components/applications/applications.test.ts
+++ b/src/components/applications/applications.test.ts
@@ -3,7 +3,7 @@ import nock from 'nock';
 import {viewApplication} from '.';
 
 import * as data from '../../lib/cf/cf.test.data';
-import anApp from '../../lib/cf/test-data/app';
+import {anApp} from '../../lib/cf/test-data/app';
 import {createTestContext} from '../app/app.test-helpers';
 import {IContext} from '../app/context';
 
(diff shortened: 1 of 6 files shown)


tlwr:
Looks good to me, I look forward to writing tests which use this pattern in the future. This will cut down on the lines of almost-duplicated static definitions.

I had a question which Andy explained to me on Slack, about why we are templating strings and returning an IApplication: why are we using strings as the return type for the builder.

The answer is because (right now) we don't want to JSON.parse everywhere; and in certain cases this would render some tests useless.

I think there is a way we can make the builder return an IApplication instead of a string, such that it makes writing tests easier, but it will require a few changes outside the scope of this PR.

:shipit:

PR #396, opened 2019-08-22
Better org tests

tlwr:
What
----

- Fix org quota tests (see commit)
- Add OrgBuilder like the AppBuilder for writing better tests
- Use the OrgBuilder everywhere

How to review
-------------

Code review

Concourse

Who can review
---------------

Not @tlwr

Maybe @andy-paine?


tlwr:
After discussing with Rich

Just using `_.merge` everywhere is better, but making a utility for pagination would be good

PR #397, opened 2019-08-27
Fix email lowercase

tlwr:
What
----

Change invite process to lowercase email addresses.

This is branched of #396 for better tests. 

How to review
-------------

Code review the last commit

Read my comment below

Who can review
---------------

Not @tlwr 

---

I'm not actually sure we can do this easily, we have some other users in our user databases which have uppercased email addresses, which means we cannot blindly lowercase all email addresses, otherwise we will email users with already existing accounts if someone types in the wrong casing.

What I think we should do is forget completely about UAA for email addresses, and do everything based on the accounts database.

So the invitation code would change from:

```
const uaaUser = await uaa.findUser(emailAddress);
```

```
const accountsUser = await accounts.getUserByEmail(emailAddress.toLowerCase())
```

We only use `uaa.findUser` in two places and they are in this codepath.

### DIFF
--- a/src/components/org-users/org-users.ts
+++ b/src/components/org-users/org-users.ts
@@ -388,13 +388,15 @@ export async function inviteUser(ctx: IContext, params: IParameters, body: objec
       logger: ctx.app.logger,
     });
 
-    const uaaUser = await uaa.findUser(values.email);
+    const emailAddress = values.email.toLowerCase();
+
+    const uaaUser = await uaa.findUser(emailAddress);
     let userGUID = uaaUser && uaaUser.id;
     let invitation: IUaaInvitation | undefined;
 
     if (!userGUID) {
       invitation = await uaa.inviteUser(
-        values.email,
+        emailAddress,
         'user_invitation',
         'https://www.cloud.service.gov.uk/next-steps?success',
       );
@@ -406,7 +408,7 @@ export async function inviteUser(ctx: IContext, params: IParameters, body: objec
 
       userGUID = invitation.userId;
 
-      await accounts.createUser(userGUID, values.email, values.email);
+      await accounts.createUser(userGUID, emailAddress, emailAddress);
     }
 
     const users = await cf.usersForOrganization(params.organizationGUID);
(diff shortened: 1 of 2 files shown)


richardTowers:
Merged #396 and rebased onto master

src/components/org-users/org-users.ts
```diff
@@ -388,13 +388,15 @@ export async function inviteUser(ctx: IContext, params: IParameters, body: objec
       logger: ctx.app.logger,
     });
 
-    const uaaUser = await uaa.findUser(values.email);
+    const emailAddress = values.email.toLowerCase();
+
+    const uaaUser = await uaa.findUser(emailAddress);
```
richardTowers: Just to confirm to myself that I wasn't misunderstanding, I checked that `findUser` already ignores case:

> Attribute operators 'eq', 'co', and 'sw' MUST perform caseIgnore matching for all string attributes
http://www.simplecloud.info/specs/draft-scim-api-01.html#query-resources

So this PR just changes the invitation address and the one in paas-accounts to be lowercase, which seems fine.

richardTowers:
I'd be happy to approve this, but I don't fully understand your comment in the PR.

> I'm not actually sure we can do this easily

Do _what_ easily exactly? Are you talking about lowercasing all the existing users in the database or something?

> otherwise we will email users with already existing accounts if someone types in the wrong casing

I think the `/Users?filter=email eq "Some.Email@..."` UAA endpoint should already ignore case - does that alter your thinking?

No opinion on switching to use paas-accounts as the source of truth, but I'd want to understand the details of why we were making that decision before going down that route.

tlwr:
I wasn't aware of UAA attempting to do the right thing, in that case then this might be enough.

---

There is a bit of implicit behaviour in us using UAA.

In order for us to use paas-accounts we would have to:

- ensure all email address / username lookups are done in paas-accounts only
- lowercase all usernames and email addresses in paas-accounts
- lowercase any and all user input pertaining to email addresses

Maybe later introduce an abstraction over the 3 stores of user data we care about in this codebase? (e.g. `PaaSUser`)

tlwr:
Yeah SCIM filtering seems to do the right thing, so we just need to make sure it is lowercased in paas-accounts

richardTowers:
If UAA is already doing the "right thing" do we even need this? Have we actually confirmed there's a bug where if you invite someone with mixed capitalisation they can't accept the invitation? Or is the whole thing a red herring.

Not saying this is necessarily a bad thing anyway, but good to understand fully before going ahead with it.

tlwr:
You're right, we might not need this anymore. This does change the behaviour of updating paas-accounts.

I'm going to close this but I think we should change the behaviour of how invitations work to use:
a) paas-accounts
b) lowercase email addresses

PR #398, opened 2019-08-27
Alter the UAA invitation URL

tlwr:
What
----

It seems the base URL for UAA is only configurable when using SAML

We use UAA as an IDP and as an SP; we use UAA on two domains:
uaa.((system_domain)) and login.((system_domain))

When we invite users we want to redirect them to login.((system_domain))
Otherwise they successfully set a password
Then they log in on the UAA subdomain
Then they are redirected to the login subdomain
But they do not have a valid session in the login subdomain cookie
So they have to log in again, which is bad user experience and confusing

Change the UAA client to redirect to login.((system_domain)) and add
some tests

How to review
-------------

Code review

1. `npm run push` to your dev env
2. Invite yourself user
3. Click the invitation in your inbox which should be `login. ...`
4. Set a password
5. Log in ONCE
6. Arrive in PaaSmin with a Terms of Service prompt

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/components/org-users/org-users.test.ts
+++ b/src/components/org-users/org-users.test.ts
@@ -654,7 +654,13 @@ describe('org-users test suite', () => {
     ;
 
     nockNotify
-      .post('/').reply(200, {notify: 'FAKE_NOTIFY_RESPONSE'})
+      .post(
+        '/',
+        (body) => {
+          const {url} = body.personalisation;
+          return url === 'https://login.system_domain/invitations/accept?code=TWQlsE3gU2';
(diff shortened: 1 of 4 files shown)


AP-Hunt:
1. ✅ npm run push to your dev env
1. ✅Invite yourself user
1. ✅Click the invitation in your inbox which should be login. 
image
1. ✅Set a password
1. ✅Log in ONCE
1. ✅Arrive in PaaSmin with a Terms of Service prompt

PR #401, opened 2019-08-28
[#168031285] Fast calculator

richardTowers:
What
----

Use a parser to evaluate the expressions from the pricing plans directly, instead of sending events to postgres via the billing API.

[pegjs](https://pegjs.org/) is a simple parser generator for JavaScript that produces fast parsers.

Our billable events are calculated by evaluating formulae like:

```
$number_of_nodes * $time_in_seconds * ($memory_in_mb/1024.0) * (0.01 / 3600)
```

Currently we do this by inserting a fake event into the database with the correct details, recomputing the billable components of _every event from the beginning of time_, and then querying the database to find out how much the fake event cost.

It would be much simpler and faster if we could just evaluate the formula directly. Since the formula aren't all that complicated it's quite simple to write a grammar that does the calculation itself. The one in this PR is heavily based on the [pegjs arithmetic example](https://github.com/pegjs/pegjs/blob/master/examples/arithmetics.pegjs), but adds support for the `ceil()` function (which is equivalent to JS's `Math.ceil()`).

Using the parser means we no longer have to call `/forecast_events` in billing, so that whole code path can be cleaned up once this lands.

I've wired up `pegjs` using webpack, which means when you run `npm run build` it will generate a parser in JavaScript and bundle this into main.js. To make the tests work I had to implement a pegjs loader for Jest, and to make TypeScript happy I had to declare a module for `*.pegjs`.

Some other advantages disadvantages from one of my commit messages:

Advantages:

* It's HUGELY faster than talking to billing
* We can remove all of the /forecast_events code from billing
* It reduces the load on the billing database
* We can test that the calculator is working properly within paas-admin
  (e.g. that it's getting the maths right)
* We get to do something that almost feels like computer science

Disadvantages:

* There's a new opportunity for bugs now that we're not using the exact
  same expression evaluator as postgres (which could cause the
  calculator to give incorrect results, if we screw this up)
* If we add any logic to the pricing plans that isn't implemented in our
  simple grammar (e.g. if we used `floor` as well as `ceil`) then the
  calculator will break until we fix the grammar.

**NOTE:** I've also hardcoded the length of a month to 30 days, so this will stop the estimates from changing every month - happy to discuss whether this is the right approach.

How to review
-------------

* Code review (feel free to come chat with me if you're not sure of the grammar - it's probably an unfamiliar language to most people)
* Check that the estimates this produces in  a dev env match the ones that billing produces (maybe test before and after a `cf push`? - bear in mind that this assumes months are 30 days long)

Who can review
---------------

Not @richardTowers

### DIFF
--- a/config/jest/pegjs.js
+++ b/config/jest/pegjs.js
@@ -1,0 +1,5 @@
+module.exports = {
+  process(src) {
+    return `const peg = require("pegjs");module.exports = peg.generate(${JSON.stringify(src)});`;
+  }
+}
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -125,6 +125,14 @@ let cfg = {
             }
           }
         ]
+      },
+      {
+        test: /\.pegjs$/,
+        use: [
+          {
+            loader: 'pegjs-loader'
+          }
+        ]
       }
     ]
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -5237,8 +5237,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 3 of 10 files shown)


richardTowers:
Comparison in my dev env:

Using the old billing code:

image

Using the new billing code:

image

richardTowers:
The latest commit should fix the currency issue, so the prices should be right now. Moving the story into review.

tlwr:
This looks good to me, I looked at the prices in Rich's dev env and the prices in my dev env and they look similar (eu-west-1 vs eu-west-2)

PR #405, opened 2019-09-04
[#167810062] Improve user management resilience

AP-Hunt:
What
----
A couple of small changes to improve how we handle the scenario when a user doesn't all 3 of:

1. A record in CF
1. A record in UAA
1. A record in paas-accounts

How to review
-------------
1. Code review
1. Deploy to your dev env
1. Test it
  - [x] Invite a user to your dev env with permissions on an org
  - [x] visit the org members page
  - [x] check you can see the user
  - [x] delete their UAA account
  - [x] leave their CF records alone
  - [x] visit the org members page
  - [x] see that you don't see the invited user in the list

Who can review
---------------
Anyone

### DIFF
--- a/src/components/org-users/org-users.test.ts
+++ b/src/components/org-users/org-users.test.ts
@@ -110,6 +110,67 @@ describe('org-users test suite', () => {
     expect(response.body).toContain('Custom-origin-3');
   });
 
+  it('should not show users who don\'t have UAA accounts', async () => {
+    nockCF
+      .get('/v2/spaces/5489e195-c42b-4e61-bf30-323c331ecc01/user_roles')
+      .reply(200, cfData.userRolesForSpace)
+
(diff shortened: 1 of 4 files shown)


tlwr:
LGTM from a code review but I haven't tested it in my dev env yet.

Thoughts about adding some `expect`ations around whether the user is displayed or not in the test? https://github.com/alphagov/paas-admin/pull/405/files#diff-1307065c0898353db545a4fe89c15535R167

I.e. it does not contain the broken user GUID

AP-Hunt:
> Thoughts about adding some `expect`ations around whether the user is displayed or not in the test? 

Good shout. Amended commit incoming.

tlwr:
LGTM did the steps and it worked, 🎉 :shipit:

PR #423, opened 2019-09-27
govuk-frontend 3.2 fixes

tlwr:
What
----

- Move the pricing link to the left side and do not capitalise it
- Use `GB` instead of `gb` everywhere
- Do not use dashes in `Platform as a Service`

See commit messages for "why"

How to review
-------------

Code review

Concourse

Who can review
---------------

@richardTowers

### DIFF
--- a/config/nunjucks.config.js
+++ b/config/nunjucks.config.js
@@ -69,7 +69,7 @@ function configure(env) {
   });
 
   env.addFilter('mbtogb', (mb) => {
-    return `${(mb / 1024).toFixed(2)}<abbr title="gigabytes">gb</abbr>`;
+    return `${(mb / 1024).toFixed(2)}<abbr title="gigabytes">GB</abbr>`;
   });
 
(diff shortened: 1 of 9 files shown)


config/nunjucks.config.js
```diff
@@ -69,7 +69,7 @@ function configure(env) {
   });
 
   env.addFilter('mbtogb', (mb) => {
-    return `${(mb / 1024).toFixed(2)}<abbr title="gigabytes">gb</abbr>`;
+    return `${(mb / 1024).toFixed(2)}<abbr title="gigabytes">GB</abbr>`;
```
richardTowers: This should be GiB technically - it's a Gibibyte - https://en.wikipedia.org/wiki/Gibibyte

tlwr:
I've reverted a dependabot PR to upgrade the nock types because it broke master

config/nunjucks.config.js
tlwr: ❇️

PR #427, opened 2019-09-27
Remove govuk-body from `<body>`

### LINKED ISSUE #426: <body> element should not have govuk-body class
Hey folks,

Just noticed that the skip link is rendered strangely in the PaaS admin app, with an unexpected bottom margin:

Screenshot 2019-09-27 at 16 39 47BST

From digging in, it looks like this is happening because you're applying the `govuk-body` class to the body

https://github.com/alphagov/paas-admin/blob/662fb6639c8957ffff8c300efc843135bc800469/src/layouts/govuk.njk#L8

I can totally understand from the name why that might have happened (!) but the [`govuk-body` class is intended to be applied to paragraphs](https://design-system.service.gov.uk/styles/typography/#paragraphs). The only class we expect to be on the `<body>` is `govuk-template__body` (which is there)

This is setting a bottom margin on the `<body>` which is being inherited by the skip link (and possibly other elements on the page) – the fact that the margin is being inherited at all is kind of weird and I'm going to raise an issue about that over on our side.

Note that this will also currently be setting the font on everything within the `<body>`, so when you remove this there's a chance you might end up with some parts of the UI not using the font, if they don't have the correct classes.

richardTowers:
What
----

The .govuk-body class is only intended for body text (like paragraphs
and so on), not the `<body>` element. Putting it on the `<body>` adds a
bunch of unintended styles. See #426.

Because we've had the class on the body all along we've got used to the
default font-family being GDS Transport, and removing it now causes a
bunch of things to fall through to the browser's default (serif,
usually). Rather than going through and adding classes to all of these
elements it's easier just to set the font-family on the body.

`.govuk-body` did a bunch of other things which will subtly change the
layouts when we remove it. Better to show than tell:

| Before | After |
|--------|------|
| Screenshot_2019-09-27 Organisations | Screenshot_2019-09-27 Organisations(1) |
| Screenshot_2019-09-27 the-system_domain-org-name – Spaces(1) | Screenshot_2019-09-27 the-system_domain-org-name – Spaces |
| Screenshot_2019-09-27 name-2064 - Overview(1) | Screenshot_2019-09-27 name-2064 - Overview |
| Screenshot_2019-09-27 name-79 - Application Overview(1) | Screenshot_2019-09-27 name-79 - Application Overview |
| Screenshot_2019-09-27 Invite a new team member(1) | Screenshot_2019-09-27 Invite a new team member |

Basically some tiny changes to margins, and some font-size changes for some of the headings.

How to review
-------------

* Code review / look at the screenshots
* If you're feeling thorough, push it to your dev env

Who can review
---------------

Not @richardTowers

Fixes #426

### DIFF
--- a/src/layouts/govuk.njk
+++ b/src/layouts/govuk.njk
@@ -5,7 +5,6 @@
 {% extends "govuk-frontend/govuk/template.njk" %}
 
 {% set htmlLang = "en" %}
-{% set bodyClasses = "govuk-body" %}
 
 {% block pageTitle %}PaaS Administration Tool{% endblock %}
 
--- a/src/layouts/govuk.screen.scss
+++ b/src/layouts/govuk.screen.scss
@@ -14,6 +14,10 @@ $govuk-global-styles: true;
 
 @import '../components/statements/statements';
 
+.govuk-template__body {
+  font-family: "GDS Transport", Arial, sans-serif;
+}
+
 .text-right {
   text-align: right;
 }


richardTowers:
Oh, this fixes the weird margin / underline on the skip link from #426 too:

image

PR #448, opened 2019-10-02
Support: Quota usage is only based on running apps

### LINKED ISSUE #445: Quota on organisations page includes stopped apps
When viewing the organisation quota page, the total running/percentage used includes stopped apps. This doesn't _actually_ stop you from staging/launching new apps so isn't as scary as it looks but it feels like a bug if I'm allowed to use >100% of my quota.

Screenshot from 2019-10-01 18-26-12 - 1

tlwr:
What
----

Quota usage is only about how many apps are running, you can have
arbitrarily many stopped apps that do not contribute to your quota.

Closes #445

How to review
-------------

Code review

Tests

Deploy to your dev env

Who can review
---------------

Not @tlwr


richardTowers:
I haven't deployed to my dev env, but I think I'm happy just with code review / seeing the tests pass.

PR #474, opened 2019-10-18
Add organisation owner to /reports/organisations

richardTowers:
What
----

I'm using annotations from the cf v3 api to record who the owner for
each organisation is. Because lables don't let you put spaces in the
values I went with annotations, but these aren't in the CLI yet so you
have to use `cf curl` to get / set them:

```
cf curl /v3/organizations/some-guid -d '{"metadata":{"annotations":{"owner":"GDS"}}}' -X PATCH

cf curl /v3/organizations/some-guid
```

So that the engagement team / people who are doing engagement can see
which organisation belongs to which organisation owner I've added a
column to the organisations report that shows the annotation.

This meant I had to switch to using the V3 API for organisations on this
page. Since this is the first V3 API we've interacted with that meant I
had to write the boilerplate to handle pagination, and the associated
tests.

I've only switched the /reports/organisations view to use V3 API, so all
the end-user facing views are still using V2 - this minimises the blast
radius if I've screwed anything up.

How to review
-------------

* Code review
* Check out this screenshot:

image

* `cf push` to a dev env and check out `/reports/organisations`
* `cf curl "/v3/organizations/$(cf org govuk-paas --guid)" -d '{"metadata":{"annotations":{"owner":"GDS"}}}' -X PATCH` to set an annotation for the govuk-paas organisation

Who can review
---------------

Not @richardtowers

### DIFF
--- a/src/components/reports/organizations.njk
+++ b/src/components/reports/organizations.njk
@@ -36,6 +36,7 @@
     <thead class="govuk-table__head">
       <tr class="govuk-table__row">
         <th class="govuk-table__header">Organisation</th>
+        <th class="govuk-table__header">Owner</th>
         <th class="govuk-table__header">Quota</th>
         <th class="govuk-table__header">Creation date</th>
         <th class="govuk-table__header">Status</th>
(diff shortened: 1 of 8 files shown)


AP-Hunt:
Should there be an accompanying PR on [paas-cf/scripts/create-org.sh](https://github.com/alphagov/paas-cf/blob/master/scripts/create-org.sh) to set this when the org is created, too?

richardTowers:
> Should there be an accompanying PR on [paas-cf/scripts/create-org.sh](https://github.com/alphagov/paas-cf/blob/master/scripts/create-org.sh) to set this when the org is created, too?

Raised https://github.com/alphagov/paas-cf/pull/2109

tlwr:
:shipit:

PR #493, opened 2019-11-01
[#169180189] Add backing service metrics for RDS services

mogds:
What
----

Initially we're using GetMetricWidgetImage as a quick way of getting metrics in
front of users. We know these are not good for accessibility, so we'll replace
them with "proper" graphs in a future story.

We're only doing RDS services (postgres/mysql) to begin with because it's easy
to work out which AWS resource corresponds to a service in cloud foundry. This
is a bit more tricky for Redis because there's not an obvious mapping between
the service instance id and the elasticache cluster id.

We've decided not to test the image generation code because it's only temporary.

How to review
-------------

Code review

Who can review
---------------

Not me or @richardTowers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,443 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@aws-sdk/abort-controller": {
+      "version": "0.1.0-preview.7",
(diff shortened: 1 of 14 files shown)


richardTowers:
Screenshots:

An RDS service:

image

Any other service:

image

tlwr:
Code review looks good

PR #495, opened 2019-11-04
[#169180189] Add service metrics (attempt 2)

richardTowers:
What
----

Add a metrics tab to the service page which shows metrics for RDS
services (postgres / mysql).

Initially we're using GetMetricWidgetImage as a quick way of getting
metrics in front of users. We know these are not good for accessibility,
so we'll replace them with "proper" graphs in a future story.

We're only doing RDS services (postgres/mysql) to begin with because
it's easy to work out which AWS resource corresponds to a service in
cloud foundry. This is a bit more tricky for Redis because there's not
an obvious mapping between the service instance id and the elasticache
cluster id.

We've decided not to test the image generation code because it's only
temporary.

This is the same as #493, but we've added a check that the user has
access to the service they're viewing metrics for and we've fixed a
hardcoded region.

How to review
-------------

* Code review
* Push it to a dev environment and manually test

Who can review
---------------

Not @richardtowers or @mogds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,443 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@aws-sdk/abort-controller": {
+      "version": "0.1.0-preview.7",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-0.1.0-preview.7.tgz",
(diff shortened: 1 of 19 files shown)


richardTowers:
Probably should also get Luke to have a look before shipping it.

richardTowers:
Screenshots:

image
image

richardTowers:
I've discussed this with Luke now - should be good to go if code review is fine.

PR #505, opened 2019-11-11
Remove unnsecessary check

paroxp:
What
----

We've discussed this with @tlwr, and believe that this isn't
essentially necessary.

There is a potential for someone locking themselves out of their
organisation or remove a billing manager, which would cause us some
pain, but frankly at the moment, this is replaced by another bug, where
inviting users have this box disabled.

This could be fixed in another way, but could mean going through a
rabbit hole, and perhaps not provide that much value to us.

We believe that if a user locks themselves out accidentally, we as
GOV.UK PaaS team, will be helpful enough to respond to their support
ticket and help them out.

How to review
-------------

- Sanity check
- See if you agree with Toby and I

Who can review
---------------

Not @tlwr


tlwr:
Needs a rebase :(

PR #508, opened 2019-11-13
Fix checkboxes

paroxp:
What
----

At the moment we're dealing with two ways of managing the permissions
via the Admin interface. This is confusing especially when it comes down
to Nunjucks and the lack of types...

Converging onto a single way of dealing with it in the typescript side,
should make it easier to follow for people.

We're fixing the bug with checkboxes being disabled, when there aren't
enough of managers in the organisation.

Closes #505

Writing some tests, to be assured that some bugs will not re-occur.

How to review
-------------

- Sanity check
- See if new tests cover enough of edge cases
- ClickOps your way for a while pretending to be a QA

### DIFF
--- a/src/components/org-users/permissions.njk
+++ b/src/components/org-users/permissions.njk
@@ -66,7 +66,7 @@
             {
               value: "1",
               html: "<span>Org manager</span>",
-              checked: values.org_roles[organization.metadata.guid].managers === '1',
+              checked: values.org_roles[organization.metadata.guid].managers == '1',
               disabled: isCheckboxDisabled
             }
           ]
@@ -87,7 +87,7 @@
             {
               value: "1",
               html: "<span>Billing manager</span>",
-              checked: values.org_roles[organization.metadata.guid].billing_managers === '1',
+              checked: values.org_roles[organization.metadata.guid].billing_managers == '1',
               disabled: isCheckboxDisabled
             }
           ]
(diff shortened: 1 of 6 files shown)


src/components/org-users/permissions.njk
```diff
@@ -163,46 +175,67 @@
...
-            items: [
-              {
-                value: "1",
-                html: "<span>Space auditor</span>",
-                checked: values.space_roles[space.metadata.guid].auditors
-              }
-            ]
-          }) }}
+          <input type="hidden" value="{{ values.space_roles[space.metadata.guid].auditors.current }}" name="space_roles[{{ space.metadata.guid }}][auditors][current]">
+          <div class="govuk-form-group">
```
richardTowers: This bit is repeated almost identically seven times - maybe worth pulling it out into a macro and just passing in the bits that change, so we'd call something like:

```
{{ roleCheckbox("org_roles", "managers", "Org manager", isCheckboxDisabled) }}
...
{{ roleCheckbox("org_roles", "billing_managers", "Billing manager", isCheckboxDisabled) }}
...
{{ roleCheckbox("org_roles", "developers", "Org developer", false) }}
...
{{ roleCheckbox("org_roles", "auditors", "Org auditor", false) }}
...
{{ roleCheckbox("space_roles", "managers", "Space manager", false) }}
...
{{ roleCheckbox("space_roles", "developers", "Space developer", false) }}
...
{{ roleCheckbox("space_roles", "auditors", "Space auditor", false) }}
```

richardTowers:
👍  looks good!

PR #522, opened 2019-11-22
Perioded metrics

paroxp:
What
----

We'd like to allow users to pick from a range of times latest to their
relative local time. For instance:

Last 1h, Last 3h, Last 24h, Last week, Last month

In order to assure this is indeed picking a range `NOW` - `OFFSET` at
all times, and not hack around with html/forms to inject ranges into the
system, we can have an additional endpoint that will be generating the
ranges for us and redirect back to the view with the set of ranges.

This for instance could have been a problem, if user has a stale view of
the page, say not refreshed for a day, and then they'd pick an offset
such as 15m. The pre-baked range would mean they're looking at 15m slot
one day ago.

This _should_ also solve the problem, of the server being too slow, and
the user manages to choose an offset and immediately attempt to update
the form with submit button.

We'd like to allow our users to choose specific range for their
representation of metrics, to help them better visualise what's going on
on our platform.

This is a rough design and should be revisited, hopefully with a
designer.

We're also reducing the size of the charts themselves, to perhaps clean
up a little visually and leave us with some room with potential
improvements to the charts themselves, for instance a tooltip on a side.

How to review
-------------

- Code review
- Sanity check
- `npm run push` and have a play around
- **Validate if the period calculation is good enough...**

### Validate if the period calculation is good enough...

I really don't like the way I solved it... I think I'd like to do something along the lines of:

```diff
diff --git a/src/components/service-metrics/service-metrics.ts b/src/components/service-metrics/service-metrics.ts
index de13960..21849a5 100644
--- a/src/components/service-metrics/service-metrics.ts
+++ b/src/components/service-metrics/service-metrics.ts
@@ -159,23 +159,16 @@ export async function viewServiceMetrics(ctx: IContext, params: IParameters): Pr
 }
 
 function parseRange(start: string, stop: string): IRange {
+  const numberOfPoints = 300;
   const rangeStart = moment.unix(parseInt(start, 10));
   const rangeStop = moment.unix(parseInt(stop, 10));
   const secondsDifference = rangeStop.diff(rangeStart) / 1000;
-  let period;
+  const period = moment.duration(Math.floor(secondsDifference / numberOfPoints), 'seconds');
 
   if (rangeStop.isBefore(rangeStart)) {
     throw new UserFriendlyError('Invalid time range provided');
   }
 
-  if (secondsDifference < 90000) { // If less than 25 hours
-    period = moment.duration(5, 'minutes');
-  } else if (secondsDifference < 1080000) { // If less than 300 hours
-    period = moment.duration(1, 'hour');
-  } else {
-    period = moment.duration(1, 'day');
-  }
-
   if (rangeStart.isBefore(rangeStop.clone().subtract(1, 'year'))) {
     throw new UserFriendlyError('Invalid time range provided. Cannot handle more than a year of metrics');
   }
```

but would love to hear @richardTowers opinions on the matter... surely some edge cases would blow up? 🤔 

## Before

Screenshot 2019-11-22 at 11 09 23

## After

Screenshot 2019-11-22 at 11 08 08
---
Screenshot 2019-11-22 at 11 08 13

### DIFF
--- a/config/nunjucks.config.js
+++ b/config/nunjucks.config.js
@@ -49,6 +49,14 @@ function configure(env) {
     return moment(date).format('HH:mm');
   });
 
+  env.addFilter('nicedatetime', (date) => {
+    return moment(date).format('h:mma, D MMMM YYYY');
+  });
+
+  env.addFilter('dateTime', (date) => {
+    return moment(date).format('YYYY-MM-DD[T]HH:mm');
+  });
+
   env.addFilter('relativetime', (date) => {
     return moment(date).fromNow();
   });
--- a/src/components/service-metrics/range-picker.njk
+++ b/src/components/service-metrics/range-picker.njk
@@ -1,0 +1,71 @@
+{% from "govuk-frontend/govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk-frontend/govuk/components/input/macro.njk" import govukInput %}
+
+<div class="paas-statement-filters">
+  <p class="govuk-body">
+    Showing metrics between <strong class="non-breaking">{{ rangeStart | nicedatetime }}</strong>
+    and <strong class="non-breaking">{{ rangeStop | nicedatetime }}</strong> for <a href="{{ linkTo(
+      'admin.organizations.spaces.services.view',
+      {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid}
+      )}}" class="govuk-link non-breaking">{{ service.entity.name }}</a>.
+  </p>
+
+  <p class="govuk-body">  
+    Each point is an average over <strong class="non-breaking">{{ period.humanize() }}</strong>.
+  </p>
+
+  <details class="govuk-details" data-module="govuk-details">
+    <summary class="govuk-details__summary">
+      <span class="govuk-details__summary-text">
+        Change time period
+      </span>
+    </summary>
+
+    <ol class="govuk-list">
+      <li>
+        <a href="{{ linkTo('admin.organizations.spaces.services.metrics.redirect', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid, offset: '1h'}) }}" class="govuk-link govuk-link--no-visited-state" title="last 1 hour">Last 1 hour</a>
+      </li>
+      <li>
+        <a href="{{ linkTo('admin.organizations.spaces.services.metrics.redirect', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid, offset: '3h'}) }}" class="govuk-link govuk-link--no-visited-state" title="last 3 hours">Last 3 hours</a>
+      </li>
+      <li>
+        <a href="{{ linkTo('admin.organizations.spaces.services.metrics.redirect', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid, offset: '12h'}) }}" class="govuk-link govuk-link--no-visited-state" title="last 12 hours">Last 12 hours</a>
+      </li>
+      <li>
+        <a href="{{ linkTo('admin.organizations.spaces.services.metrics.redirect', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid, offset: '24h'}) }}" class="govuk-link govuk-link--no-visited-state" title="last 24 hours">Last 24 hours</a>
+      </li>
+      <li>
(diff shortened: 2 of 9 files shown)


richardTowers:
Screenshot of the interface post design surgery:

image

config/nunjucks.config.js
```diff
@@ -49,6 +49,14 @@ function configure(env) {
     return moment(date).format('HH:mm');
   });
 
+  env.addFilter('nicedatetime', (date) => {
```
richardTowers: I think the GO.UK styleguide is like 4 June 2017: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

paroxp: noice

src/components/service-metrics/range-picker.njk
```diff
@@ -0,0 +1,73 @@
+{% from "govuk-frontend/govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk-frontend/govuk/components/input/macro.njk" import govukInput %}
+
+<div class="paas-statement-filters">
+  <p class="govuk-body">
```
richardTowers: Discussed with Luke and we reckon some copy like:

> Showing metrics between **1am 23 October 2019** and **0am 22 November 2019** for [name-1508](#).
> 
> Each point is an average over a day.

src/components/service-metrics/range-picker.njk
```diff
@@ -0,0 +1,73 @@
...
+  <details class="govuk-details" data-module="govuk-details">
+    <summary class="govuk-details__summary">
+      <span class="govuk-details__summary-text">
+        Change time period
+      </span>
+    </summary>
+
+    <ol class="govuk-list">
+      <li>
+        <a href="{{ linkTo('admin.organizations.spaces.services.metrics.redirect', {organizationGUID: organization.metadata.guid, spaceGUID: space.metadata.guid, serviceGUID: service.metadata.guid, offset: '1h'}) }}" class="govuk-list govuk-link--no-visited-state" title="last 1 hour">Last 1 hour</a>
```
richardTowers: Should these have `govuk-list` or `govuk-link` 🤔

src/components/service-metrics/service-metrics.ts
```diff
@@ -8,10 +8,41 @@ import { IContext } from '../app';
...
+  const rangeStop = moment();
+  const timeRanges: {[key: string]: moment.Moment} = {
+    '1h': rangeStop.clone().subtract(1, 'hour'),
+    '3h': rangeStop.clone().subtract(3, 'hours'),
+    '12h': rangeStop.clone().subtract(12, 'hours'),
+    '24h': rangeStop.clone().subtract(24, 'hours'),
+    '7d': rangeStop.clone().subtract(7, 'days'),
+    '30d': rangeStop.clone().subtract(30, 'days'),
+  };
+  const rangeStart = timeRanges[params.offset] || timeRanges['1h'];
```
richardTowers: I think the default should be `24h`, rather than `1h` (at least, that's what it was before).

src/components/service-metrics/service-metrics.ts
```diff
@@ -25,6 +56,21 @@ export async function viewServiceMetrics(ctx: IContext, params: IParameters): Pr
...
 
+    if (!params.rangeStart || !params.rangeStop) {
+      return {
+        status: 302,
+        redirect: ctx.linkTo(
+          'admin.organizations.spaces.services.metrics.redirect', {
+            organizationGUID: params.organizationGUID,
+            spaceGUID: params.spaceGUID,
+            serviceGUID: params.serviceGUID,
+            offset: '1h',
```
richardTowers:
```suggestion
            offset: '24h',
```

richardTowers: Perhaps we could save ourselves one redirect bounce here too, and redirect to:

```
ctx.linkTo('admin.organizations.spaces.services.metrics.view', {
      organizationGUID: params.organizationGUID,
      spaceGUID: params.spaceGUID,
      serviceGUID: params.serviceGUID,
      rangeStart: moment().subtract(24, 'hours').unix(),
      rangeStop: moment().unix(),
    })
```

src/components/service-metrics/range-picker.njk
paroxp: Not certain what you mean. They do, the `govuk-list` is on the `ol` element and the `govuk-link` is on the `a` element.

As per: https://design-system.service.gov.uk/styles/typography/#lists and https://design-system.service.gov.uk/styles/typography/#links

Do you perhaps consider removing these?

richardTowers: This one is `<a ... class="govuk-list...>` though, which looks wrong?

paroxp: ah, beautiful!

PR #523, opened 2019-11-22
Add a platform admin page

tlwr:
What
----

A picture is worth a thousand lines of code:

image

_An image of the platform admin page_

Platform admin is an abstraction borrowed from Notify, that I have repurposed to represent global auditors, admins, and read-only admins.

This PR adds a platform admin page to help us find all the secret features that platform admins have, this will be especially useful given that we have global auditors for engagement and PMO teams.

During this I also added a 'not authorised' middleware so we can stop returning 404s everywhere

How to review
-------------

- [x] Code review

- [x] Look at the tests / linter

- [x] Deploy to your dev environment

- [x] Attempt to use the feature, as a non-platform admin

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/components/app/app.ts
+++ b/src/components/app/app.ts
@@ -8,7 +8,7 @@ import helmet from 'helmet';
 import { IncomingMessage, ServerResponse } from 'http';
 import { BaseLogger } from 'pino';
 
-import { IResponse } from '../../lib/router';
+import { IResponse, NotAuthorisedError } from '../../lib/router';
 import auth from '../auth';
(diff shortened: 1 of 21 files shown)


src/components/app/app.ts
```diff
@@ -90,6 +90,10 @@ export default function(config: IAppConfig) {
 
   app.get('/healthcheck', (_req: express.Request, res: express.Response) => res.send({message: 'OK'}));
 
+  app.get('/forbidden', () => {
```
paroxp: This seems to only be used by tests :(

PR #526, opened 2019-11-25
Events pagination QoL and platform admin QoL

tlwr:
What
----

Quality of life changes for event pagination and platform admin

When we're looking through events, currently the pagination is only at the top of the page. In order to move to the next page, one must scroll back up. Instead we should have identical pagination at the top and bottom, when there are multiple pages. This is done using a set block.

When we're on the platform admin page, we should default the costs year field to be this year.

How to review
-------------

Code review

Who can review
---------------

Not @tlwr

### DIFF
--- a/src/components/platform-admin/homepage-costs.njk
+++ b/src/components/platform-admin/homepage-costs.njk
@@ -43,7 +43,8 @@
                  id="view-costs-year"
                  name="year"
                  type="number"
-                 pattern="20[123][0-9]">
+                 value="{{ thisYear }}"
+                 pattern="20[123][0-9]"/>
         </div>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


src/components/platform-admin/homepage-costs.njk
```diff
@@ -43,7 +43,8 @@
                  id="view-costs-year"
                  name="year"
                  type="number"
-                 pattern="20[123][0-9]">
+                 value="{{ thisYear }}"
+                 pattern="20[123][0-9]"/>
```
richardTowers: gov.uk paas end date confirmed

tlwr:
I did start to full it out into a separate set of templates, however it ended up looking quite gross:

```
pagination_controls(next_link, previous_link, page, total_pages)
```

but then arithmetic had to be done in the next link, or string concatenation for the page link

and I wasn't happy with it

PR #529, opened 2019-11-26
Charts summary table

paroxp:
What
----

There is a lot of data and we're planning on bringing in more. It's
difficult to manage all different divs across two files. Totally
unreasonable if we can make it a macro and share the contents.

This way, we can modify a single HTML block that will fit all purposes.

The charts are great! And a good start for the bright future of
providing our valued tenants with some out of the box metrics.

We can make improvements to these charts, by collecting some data and
displaying them to the tenant aside of the chart.

We believe, fields like `Current`, `Average`, `Min` and `Max` will be
really useful to the viewer.

This would potentially allow us to make the charts interactive in the
future.

How to review
-------------

- Sanity check
- See TLWR's environment
- Deploy it to your env

### DIFF
--- a/config/nunjucks.config.js
+++ b/config/nunjucks.config.js
@@ -3,6 +3,13 @@ const jmespath = require('jmespath');
 const showdown = require('showdown');
 const moment = require('moment');
 
+const longUnits = {
+  KiB : 'kibibytes',
+  MiB : 'mibibytes',
+  GiB : 'gibibytes',
+  TiB : 'tebibytes',
+};
+
 function configure(env) {
   env.addFilter('query', (data, query) => {
     return jmespath.search(data, query);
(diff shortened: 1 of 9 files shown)


AP-Hunt:
First thoughts: can we convert the values in the tables in to more human-friendly units? E.g. kb/mb/gb?

paroxp:
Absolutely :) Glad you noticed that whilst I'm working on it.

lukemalchergds:
Looks good. Some Ideas:
- Align top of summary table to top of the chart 
- ensure table captions are consistent with chart titles
- change title to "summary of [chart title]" OR to balance accessibility but prevent repetition I'd be inclined to simply label it "Summary", but have some screen-reader only text appended saying "of [chart title]"
- I dont think this needs visible column headers, but have we marked up the table in a way we think is accessible? reference: https://www.w3.org/WAI/tutorials/tables/ ?
- When the elements on the page stack, the spacing between the elements on the page is a bit off. The summary table for a chart should be closest to the chart it is associated with.
Screen Shot 2019-11-26 at 17 43 58

PR #532, opened 2019-11-28
Unify declaration of links

paroxp:
What
----

A lot of links across the board don't implement the `govuk-link` class.
Which in theory is not an issue, but means we're not truly using the
design system and there may be edge cases that Design System covers
where we forget about them. Let's trust our friends from the Design
System team!

Rafal is also going mad with the ":visited" variants of links which
aren't content but more of a navigation function. It appears odd, that
you only ever visit one of say three of the spaces on the list. Adding
the `govuk-link--no-visited-state` will provide us with the eye-candy
and potentially better navigation around the system, and will make Rafal
happy!

How to review
-------------

- Sanity check
- Make sure the `govuk-link--no-visited-state` class is added to sensible links and not content 🤔 (assuming you are on the same wave lengths with Rafal)

### DIFF
--- a/src/components/account/access-denied.njk
+++ b/src/components/account/access-denied.njk
@@ -17,7 +17,7 @@
         If you have any queries, contact
         <a class="govuk-link" href="https://www.cloud.service.gov.uk/support">support</a>.
       </p>
-      <a href="{{ linkTo("admin.home") }}">Back to account</a>
+      <a href="{{ linkTo("admin.home") }}" class="govuk-back-link">Back to account</a>
     </div>
   </div>
 {% endblock %}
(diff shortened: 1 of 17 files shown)


tlwr:
For the visited case, would it be more sensible (as this is an admin app) to do a CSS override such that any non-external link has the normal colour when visited:

```
a.govuk-link[href^="http://"]:visited,
a.govuk-link[href^="https://"]:visited {
  color: $govuk-link-colour;
}
```

paroxp:
I don't think so. I think there will be occasions where I'd like to have the visited state on. For instance:

> [View our guide](https://docs.cloud.service.gov.uk/get_started.html#set-up-the-cloud-foundry-command-line) on logging in and deploying your first app to GOV.UK PaaS.

Link to documentation like this one would benefit for an individual to think "Oh yeah, I've been here".

What do you think?

PR #558, opened 2020-01-02
[#169504332] Download metrics data as a CSV

mogds:
What
----

This adds a link below each chart to download redis and rds backing service
metrics as a CSV. This is so we can provide an alternative for assistive
technology users.

How to review
-------------

Code review

Who can review
---------------

Not me


tlwr:
Rebased this onto master and added CSV download for CDN routes

PR #559, opened 2020-01-02
Improve the Y-axis formatting of metrics with big values

AP-Hunt:
What
----

Some of the metrics we plot for tenants can have big values, and big ranges of
values, often spanning tens of hundreds of thousands.

Initially we used the `,.2r` D3 format string, which produces grouped thousands
with two significant digits [1], such as "150,000". Unfortunately, with larger
values the resulting string would be too big for the space on the left of the
Y-axis, and would get chopped off a bit.

Now we use the `.2s` D3 format string [1] for metrics we know can have big
values. This format produces strings with an SI-prefix and 2 significant
digits. For example, the value 42,000,000 is formatted as "42M". This should
make the strings shorter, and more likely to fit in space available.

[1] https://github.com/d3/d3-format/tree/v1.3.1#locale_format

How to review
-------------
1. Code review
1. Check I haven't missed any metrics which could do with the `.2s` format.

Who can review
---------------
Not I


tlwr:
This is my fault but could we make `mTotalErrorRate` consistent with `m{4,5}xx....` for CloudFront, and them all by `s` not `r`?

AP-Hunt:
> This is my fault but could we make `mTotalErrorRate` consistent with `m{4,5}xx....` for CloudFront, and them all by `s` not `r`?

Done :+1:

PR #586, opened 2020-01-20
Migrate templating to Server Side React

paroxp:
What
----

The time has come, to retire nunjuck in our system. This is based on the fact that it actively has no types or type checking as well as it being really difficult to configure.

We've made some decisions through out this refactor to make it easier for ourselves, but also standard enough for us to be able to sed it out for something else. These are:

- Use TSX
- Use mostly functional React
- Name convention: `Component(props: IComponentProperties)` and `ComponentPage(props: IComponentPageProperties)`
- Moving all template files into `views.tsx`
- Moving all controllers into `controllers.tsx`
- Separate text and html node elements with `{' '}` and test for that being a case

How to review
-------------

This is a rather big change... Recommend going commit by commit.

I'd like to ask for particular attention to:

- `Setup a templating engine` - simplest thing to make it work I guess...
- `Add space separation testing` - it's a shame, but it has to be done.
- `Reactify Spaces component` - was a big one 😱 
- `Reactify Reports component` - change in CSV of the chats generation, with a no longer existing trailing new line
- `Reactify the org-users component` - was a big one 😱 I'd like the reviewer to take special interest in permission settings and user invites...
- `Reactify Calculator component` - was a big one 😱 
- `Reactify Service Metrics component` - compressed multi-view to single-view system, there is also change in CSV of the chats generation, with a no longer existing trailing new line


tlwr:
_Dropping comments below as I review_

This is a large enough PR that probably warrants multiple reviews

tlwr:
[Enzyme](https://airbnb.io/enzyme/) is a React testing library maintained by Airbnb and has support for shallow rendering. It being used in the tests looks nice and ergonomic

paroxp:
Agree, sadly for "let's not get blocked" reasons, I've mixed and matched cheerio with enzyme.

src/layouts/react-spacing.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const regexTextAfterElement = /\<\/[0-9a-zA-Z]\>[0-9a-zA-Z\(\[]+/g;
+
+  const missingSpacesBefore = body.match(regexTextBeforeElement) || [];
+  const missingSpacesAfter = body.match(regexTextAfterElement) || [];
+
+  return [...missingSpacesBefore, ...missingSpacesAfter];
+}
+
+describe(testSpacing, () => {
+  it('does the right thing', () => {
```
tlwr:
```suggestion
  it('correctly checks whether any spaces are missing before or after elements', () => {
```

src/components/account/account.test.tsx
```diff
@@ -93,13 +94,14 @@ describe('account test suite', () => {
 
         const response = await account.getGoogleOIDCCallback(ctx, params);
         expect(response.body).toContain('Access Denied');
+        expect(testSpacing(response.body as string)).toHaveLength(0);
```
46bit: These two lines seem incompatible. How can `response.body` include *Access Denied* and yet be empty?

src/components/account/account.test.tsx
```diff
@@ -108,36 +110,36 @@ describe('account test suite', () => {
 
         const response = await account.getGoogleOIDCCallback(ctx, params);
         expect(response.body).toContain('try again later');
+        expect(testSpacing(response.body as string)).toHaveLength(0);
```
46bit: Ditto with above. I'm guessing this is possible then. It may help to rename `testSpacing` to describe its behaviour.

P.S., I'm OK with this being done in a followup PR. 🤗

src/components/application-events/controllers.test.tsx
```diff
@@ -226,8 +230,8 @@ describe('application events', () => {
 
       expect(response.body).toContain('Displaying page 1 of 2702');
       expect(response.body).toContain('1337 total events');
-      expect(response.body).toContain('<a class="govuk-link" disabled>Previous page</a>');
-      expect(response.body).not.toContain('<a class="govuk-link" disabled>Next page</a>');
```
46bit: Both these lines removed the closing `"` along with `disabled`.

src/components/application-events/views.test.tsx
```diff
@@ -0,0 +1,52 @@
+import cheerio from 'cheerio';
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import { IApplication, IAuditEvent } from '../../lib/cf/types';
+import { ApplicationEventsPage } from './views';
+
+describe(ApplicationEventsPage, () => {
+  const application = {metadata: {guid: 'APPLICATION_GUID'}, entity: {name: 'test-app'}} as unknown as IApplication;
```
46bit: I'm curious about this bit of TypeScript. Does `{ … } as unknown as Type` typecheck where `{ … } as Type` would not?

src/components/errors/middleware.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+          name: 'admin.organizations',
+          path: '/',
+        },
+      ]);
+
+      req.csrfToken = () => '';
+
+      next();
+    });
+  });
```
46bit: Could you describe what is important or what the failure mode might be?

src/components/account/account.test.tsx
paroxp: This response is wrapped in the `testSpacing` function, which returns an array list. If the list is not empty, it should throw an error.

It's purpose is to check the HTML content and establish if cases[with no spacing]()occur for the end user.

src/layouts/react-spacing.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const regexTextAfterElement = /\<\/[0-9a-zA-Z]\>[0-9a-zA-Z\(\[]+/g;
+
+  const missingSpacesBefore = body.match(regexTextBeforeElement) || [];
+  const missingSpacesAfter = body.match(regexTextAfterElement) || [];
+
+  return [...missingSpacesBefore, ...missingSpacesAfter];
+}
+
+describe(testSpacing, () => {
+  it('correctly checks whether any spaces are missing before or after elements', () => {
```
46bit: I suggest renaming `testSpacing` to `spacesMissingAroundInlineElements` or similar? :)

src/layouts/partials.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const markup = shallow(<Header location="London" isPlatformAdmin={false} />);
+    expect(markup.find('header nav li a.paas-govuk-tag').text()).toMatch('London');
+    expect(markup.find('header nav li.admin')).toHaveLength(0);
+    // The following is for simply compliance with the design system.
+    // https://github.com/alphagov/govuk-frontend/issues/1688
+    expect(markup.find('header.govuk-header .govuk-header__logotype svg image')
+      .filterWhere(item => item.prop('src') === '/assets/images/govuk-logotype-crown.png')).toHaveLength(1);
+  });
+
+  it('should show the admin link if platform admin', () => {
```
46bit: An inverse test *should not show the admin link if not platform admin* would be lovely too. 💚

src/components/application-events/views.test.tsx
paroxp: This is to do with the linter more than a compiler.

Imagine a scenario, where your interface requires 5 parameters, but you are certain that the function using the interface is only using 1 parameter.

You don't necessarily want to compose the entire object (considering CF API), just to check if the `name` parameter is displayed correctly.

Linter will tell complain, that your custom object, is missing the remaining 4 parameters.

If you cast it to unknown first, and then to the type, it gets to the state where it forgets that the object is missing pieces inside...

It's a terrible hack... Should not be used in production code...

src/layouts/partials.tsx
```diff
@@ -0,0 +1,187 @@
...
+    </li>
+  );
+
+  const assetPath = params.assetPath || '/assets';
+
+  return (
+    <header className="govuk-header" role="banner" data-module="govuk-header">
+      <div className="govuk-header__container govuk-width-container">
+        <div className="govuk-header__logo">
+          <a href="#" className="govuk-header__link govuk-header__link--homepage">
```
46bit: I'm curious if the `href="#"` is intentional?

src/layouts/partials.test.tsx
paroxp: I was lazy checking for that in the previous test on line 10, without description. "Two birds with one stone" kind of thing...

I'm happy to separate it into it's own block if it'd make it simpler to read/find? :)

src/layouts/partials.tsx
paroxp: Good catch. I'd much rather have `/` there, which would redirect to admin home page... This must be an artefact of me copying design system fresh to make sure the element is roughly up to date.

src/layouts/react-spacing.test.tsx
paroxp: Sounds good :)

src/components/service-events/controllers.test.tsx
```diff
@@ -218,8 +225,8 @@ describe('service events', () => {
 
       expect(response.body).toContain('Displaying page 1 of 2702');
       expect(response.body).toContain('1337 total events');
-      expect(response.body).toContain('<a class="govuk-link" disabled>Previous page</a>');
-      expect(response.body).not.toContain('<a class="govuk-link" disabled>Next page</a>');
+      expect(response.body).toContain('<a class=\"govuk-link\">Previous page</a>');
+      expect(response.body).not.toContain('<a class="govuk-link">Next page</a>');
```
46bit: If the tests passed I guess it doesn't matter these two lines inconsistently escape double quotes 🤷‍♀

src/components/events/views.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const $ = cheerio.load(markup.html());
+    expect($('.actor dd').text()).toContain(actor.email);
+    expect($('.actor dd').text()).toContain('jeff');
+    expect(testSpacing($.html())).toHaveLength(0);
+  });
+});
+
+describe(TargetedEvent, () => {
+  const actor = {email: 'jeff@jefferson.com'} as unknown as IAccountsUser;
+  const target = {guid: 'TAGET_GUID'} as unknown as IAuditEventActorTarget;
```
46bit:
```suggestion
  const target = {guid: 'TARGET_GUID'} as unknown as IAuditEventActorTarget;
```

src/components/service-metrics/controllers.test.tsx
```diff
@@ -362,7 +378,7 @@ describe('service metrics test suite', () => {
       .filter(line => line.length > 0)
     ;
 
-    expect(lines.length).toBeGreaterThan(2);
+    expect(lines.length).toEqual(2);
     expect(lines.length).toBeLessThan(400);
```
46bit:
```suggestion
```

src/components/service-metrics/controllers.test.tsx
```diff
@@ -408,7 +424,7 @@ describe('service metrics test suite', () => {
 
     const lines = response.download.data.split('\n');
 
-    expect(lines.length).toBeGreaterThan(2);
+    expect(lines.length).toEqual(2);
     expect(lines.length).toBeLessThan(450);
```
46bit:
```suggestion
```

src/components/service-metrics/controllers.test.tsx
```diff
@@ -459,7 +475,7 @@ describe('service metrics test suite', () => {
...
```
46bit:
```suggestion
```

src/components/service-metrics/controllers.test.tsx
```diff
@@ -506,7 +522,7 @@ describe('service metrics test suite', () => {
...
```
46bit:
```suggestion
```

src/components/application-events/views.test.tsx
46bit: Oh I see! Thanks Rafal. 😄 I think that's OK in ordinary production code, but maybe not in scary places.

src/components/account/account.test.tsx
```diff
@@ -94,7 +94,7 @@ describe('account test suite', () => {
 
         const response = await account.getGoogleOIDCCallback(ctx, params);
         expect(response.body).toContain('Access Denied');
-        expect(testSpacing(response.body as string)).toHaveLength(0);
+        expect(spacesMissingAroundInlineElements(response.body as string)).toHaveLength(0);
```
46bit: This does look wonderfully better ✨

src/components/applications/views.tsx
```diff
@@ -0,0 +1,175 @@
...
+      <section className="govuk-tabs__panel">
+        {props.children}
+      </section>
+    </div>
+  </>);
+}
+
+export function AppLink(props: IAppLinkProperties): ReactElement {
+  if (props.href.match(/apps[.]internal/)) {
+    return (<>{props.href}<br /></>);
```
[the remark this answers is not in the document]
paroxp: We have been using them before, but in a shape of `<br>` - https://github.com/alphagov/paas-admin/blob/13b57503cc8b8f7cb3bd585a9d88bc66813d6958/src/components/application-events/application-events.njk#L56 for instance.

I agree that we should come back to this at some point and CSS it out :)

src/components/calculator/views.tsx
```diff
@@ -0,0 +1,241 @@
...
+                  {plans.map(plan =>
+                    <option key={plan.planGUID} value={plan.planGUID}>{plan.planName}</option>)}
+                </select>
+              </div>
+              :
+              <>
+                <input type="hidden" name={`items[${props.state.items.length}][planGUID]`} value={plans[0].planGUID} />
+                {props.serviceName === 'app' ?
+                  <>
+                    <div style={{width: '46%'}}>
```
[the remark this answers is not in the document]
paroxp: It was there previously see: https://github.com/alphagov/paas-admin/blob/f89414f3304c13fe133963e70541b225b919db5d/src/components/calculator/calculator.njk#L70

Also agree, we should perhaps re-visit this entire calculator view anyway :(

src/components/errors/middleware.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return region;
+  }
+}
+
+export function internalServerErrorMiddleware(err: Error, req: any, res: express.Response, next: express.NextFunction) {
+  req.log.error(err);
+
+  if (err instanceof NotFoundError) {
+    return pageNotFoundMiddleware(req, res, next);
+  }
```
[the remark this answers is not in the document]
paroxp: Yas. I like that!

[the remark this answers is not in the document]
paroxp: But also, agreed IRL that because it has been like this in the past, we may want not to address it in this PR.

src/components/reports/views.tsx
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  public render(): ReactElement {
+    return this.SankeyInputJSON();
+  }
+
+  private SankeyInputJSON(): ReactElement {
+    return (
+      <script id="data" type="application/json" dangerouslySetInnerHTML={{
+        __html: JSON.stringify(this.props.data),
```
richardTowers: This was probably broken before as well, but after a discussion with Rafal we spotted that there's an XSS vulnerability here.

`data` includes organisation "owners", which are based on the organisation metadata which is writeable by org managers. Someone could set the metadata to something like `</script><script>alert('haha pwned')</script>`. We'd be saved by our CSP in modern browsers, but still.

We could avoid the `dangerouslySetInnerHTML` entirely if we base64 encoded the JSON, but we'd need to teach the client side JS to base64 decode it again.

Alternatively we could escape the `/` in `</script>` which I think is a sufficient mitigation.

src/components/reports/views.tsx
```diff
@@ -0,0 +1,461 @@
...
+export function CostReport(props: ICostReportProperties): ReactElement {
+  return(<>
+    <h1 className="govuk-heading-l">
+      Billables for {props.date}
+    </h1>
+
+    <div className="govuk-grid-row">
+      <div className="govuk-grid-column-one-quarter">
+        <h2 className="govuk-heading-m">
+          {JSON.stringify(props.billableEventCount)}
```
richardTowers: Using `JSON.stringify` on a number 🤔

src/components/reports/views.tsx
paroxp: added `.replace('/', '\/')` for good measure...

src/components/reports/views.tsx
paroxp: It's a nice artefact from the past `{{ billableEventCount | dump }}` :D

Changing.

src/layouts/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const DATE = 'MMMM Do YYYY';
+export const TIME = 'HH:mm';
+export const DATE_TIME = 'h:mma, D MMMM YYYY';
+
+export const KILOBYTE =  1024;
```
richardTowers: Technically this is a KIBIBYTE

https://en.wikipedia.org/wiki/Kibibyte

richardTowers:
image

Code review done! 🎉 

I just want to do some clicky-clicky manual testing, then this is :shipit: ready

paroxp:
Did the rebase, to squash commits. Now, all of you can review again :trollface:

PR #597, opened 2020-01-21
[#170800032] Fix the instance label

paroxp:
What
----

At the moment, the regex sniffing out the instance number is not taking
an effect which is essentially causing the default `001` to kick in.

This is not that great from the user perspective as misleading and
repeated information is being presented.

We're adjusting the Regex allowing us to correctly get the number out of
the label.

As I'm not very good at regex, I've written a test case that will help
establish issues like that in the future...

How to review
-------------

- Deploy to your env
- Create HA redis
- View its metrics page and see summary table
- Expect `001` and `002`

### DIFF
--- a/src/components/service-metrics/controllers.tsx
+++ b/src/components/service-metrics/controllers.tsx
@@ -35,6 +35,8 @@ import {
   UnsupportedServiceMetricsPage,
 } from './views';
 
+const PERSISTANCE_MESSAGE_370_DAYS = '370 days';
+
 interface IRange {
   readonly period: moment.Duration;
   readonly rangeStart: moment.Moment;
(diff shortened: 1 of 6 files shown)


src/components/service-metrics/controllers.tsx
```diff
@@ -35,6 +35,8 @@ import {
   UnsupportedServiceMetricsPage,
 } from './views';
 
+const PERSISTANCE_MESSAGE_15_DAYS = '15 days';
```
tlwr: We persist metrics for up to `370d`

https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/operations.d/500-prometheus.yml#L57

src/components/service-metrics/views.tsx
```diff
@@ -176,6 +178,12 @@ function RangePicker(props: IRangePickerProperties): ReactElement {
         Each point is an average over <strong className="non-breaking">{props.period.humanize()}</strong>.
       </p>
 
+      {props.persistancePeriod
+        ? <p className="govuk-body">
+            These metrics have only <strong>{props.persistancePeriod}</strong> of persistance.
```
tlwr:
```suggestion
            These metrics are retained for up to <strong>{props.persistancePeriod}</strong>. While metrics are experimental, we cannot guarantee a minimum metrics retention period.
```

PR #615, opened 2020-01-29
Deprecate TSLint

paroxp:
What
----

The tool is being deprecated in favour of ESLint and TypeScript plugins.

Getting ahead of the curve and removing it before it's too late.

Now that we've deprecated the TSLint tool, we don't want to be in the
position where we have nothing linting the code...

How to review
-------------

- Sanity check
- Review with `&w=1`...


richardTowers:
Rafal is clearly just trying to take toby's "most lines changed in paas-admin" crown 🤔

Neither of you will be able to defeat the mighty @GovUK-PaaS-CI-User.

paroxp:
See me try! :trollface:

PR #634, opened 2020-02-04
CLI Alternative - component

paroxp:
What
----

We're using the pattern in two places, and would like to use it more.

Making it as simple for us to do so, will encourage future usage as well as progress with a story in our backlog.

How to review
-------------

- Sanity check


src/layouts/partials.tsx
```diff
@@ -249,3 +253,27 @@ export function SubNavigation(props: ISubNavigationProperties): ReactElement {
     </div>
   );
 }
+
+export function CommandLineAlternative(props: ICommandLineAlternativeProperties): ReactElement {
+  return (
+    <>
+      <h4 className="govuk-heading-s">On the commandline</h4>
+
+      <p>
```
tlwr:
```
      <p>
        You can also view the same information on the commandline, to see
        details {optional property for additional context} use:
      </p>
```

To allow us to customise the error a bit more?

paroxp: Good shout. Amended.

PR #637, opened 2020-02-05
Enhance the Graphs

paroxp:
What
----

There are two issues we observed with the current setup.

1. Large numbers (such as 10000000000000), will go off screen and not
  present itself properly on the chart
2. Some labels often repeat themselves... (10, 20, 30, 40, 50, 50, 50)

During the brief investigation, where the stub data was reproducing the
issue number 2, it came apparent that the d3Format is somewhat faulty.

This came really apparent, when we enabled the default labels for Grid
and have formatted the axis. This caused two values being printed... One
with the incorrect behaviour, and one with correct behaviour.

I believe this is related to the actual format that we have been
passing, for instance I can still reproduce it with `d3Format('.1r')`.

We're creating our own formatters and setting these up in the metrics so
we have control over our own destiny.

Additionally as a bonus...

Personally I find it difficult to look at and follow. I may be wrong,
and I haven't actually discuss this with anyone else. Not certain why
the grid was previously not a thing nor why the left axis line was not
visible.

I'm also adding some headroom to the charts, which will allow the spike
not hit the top border but always stay inside the graph.

- Re-order rendering positions

My OCD when trying to look at the graph from source code is killing me a
little.

By moving those around, we're rendering left to right, top to bottom
pretty much, with an exception for the legend labels.

How to review
-------------

- Sanity check
- Quality Control :trollface:
- Run it yourself
- Keep on refreshing - observe no issues

Screenshots
------------

Graph with the label repeated three times
Graph with the label repeated three times

Graph with the fixed label formatting and new grid
Graph with the fixed label formatting and new grid

Accidental discovery of label repeating bug
Accidental discovery of label repeating bug

### DIFF
--- a/src/components/service-metrics/metrics.tsx
+++ b/src/components/service-metrics/metrics.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { IMetricSerie, IMetricSerieSummary } from '../../lib/metrics';
 import { drawLineGraph } from '../charts/line-graph';
 import { IMetricProperties } from './views';
+import { bytesConvert } from '../../layouts';
 
 interface ISeries {
   readonly [key: string]: ReadonlyArray<IMetricSerie>;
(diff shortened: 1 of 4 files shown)


src/components/service-metrics/metrics.tsx
```diff
@@ -12,6 +13,28 @@ interface ISummaries {
...
+  const bytes = bytesConvert(value);
+
+  return `${bytes.value}${bytes.short}`;
+}
+
+export function percentLabel(value: number, _index: number): string {
+  return `${value}%`;
+}
+
+export function numberLabel(value: number, _index: number): string {
```
AP-Hunt: I think we should expand the suffix options out here to cover billions

paroxp: Fair play. Updated.

You also made me realise that the graphs may potentially want to operate on `1.1`, `1.2`, `1.3` kinda thing. Added `.toFixed(2)` for good measure.

tlwr:
This looks 🎩 

How do you think it would look with a final tick line for both the X and Y axes, so the ticks don't "poke out" - I'm not sure how easy this is, or if it would look good

paroxp:
Fair, I was previously hiding this, but it does look pretty cool.

Screenshot 2020-02-05 at 17 25 45

PR #650, opened 2020-02-11
Fix the menu handle on mobile devices

paroxp:
What
----

At the moment, this particular functionality is broken and prevents
administrators using the tool on mobile devices.

We're adding fixes to make sure the menu is mobile-first. This means, we
cannot just extend the `govuk-tag` class and have to set the values
manually.

We're also making some accessibility changes to the region link, to
indicate more precisely, what it's doing.

How to review
-------------

- Run locally
- Make sure the menu continues to be usable
- Resize your window pretending you're on mobile device, and attempt to use menu

### DIFF
--- a/src/layouts/_header.scss
+++ b/src/layouts/_header.scss
@@ -1,35 +1,68 @@
 a.paas-govuk-tag {
-  @extend .govuk-tag;
+  @include govuk-media-query(tablet) {
+    display: inline-block;
 
-  &:visited, &:hover, &:link {
-    color: #fff;
-  }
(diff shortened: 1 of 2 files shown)


tlwr:
I looked at rafalp's Pazmin and it seemed to not work when my Chrome was in responsive mode?

Would you mind showing me a demo

paroxp:
I didn't deploy it to my dev. Ran on local. I shall deploy :)

tlwr:
LGTM in Rafal's dev env

PR #686, opened 2020-03-03
Create org

paroxp:
What
----

We'd like to be deprecating more custom scripts from our repos and move
operational "one time" things to pazmin.

This functionality will allow us to create new organisations that
include the sandbox and tag it with an owner used in different section
of the system.

It will not invite users by default.

Potential improvements could be made around the "Owner" box, as we most
likely have a list of "owners" that we could display instead.

That will however do for now.

This implementation also uses v3 API which means we are only making one
call to create organisation and one call to create space, where as
script performs 7 calls all together which separately tag the
organisation and manage user admin.

How to review
-------------

- Code review
- Deploy to your env (or see rafalp)
- As platform operator, navigate to Patform Admin > Create organisation
- Play around with a form
- Establish if this is good enough

### DIFF
--- a/src/components/app/app.test-helpers.ts
+++ b/src/components/app/app.test-helpers.ts
@@ -39,7 +39,7 @@ export function createTestContext(ctx?: {}): IContext {
       ['secret'],
     ),
     viewContext: {
-      csrf: '',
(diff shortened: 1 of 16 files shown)


schmie:
:+1: LGTM

PR #693, opened 2020-03-05
Update the way we handle (non)visited link text colour

kr8n3r:
## What
Update the (non)visited link text colour handling.

Update the way we handle (non) visited links colour, by making all links not show a visited state.

As this is an admin interface there is no need to distinguish between visited and non-visited links.

govuk-frontend has both a css class `.govuk-link--no-visited-state` and a [mixin](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/helpers/_links.scss#L146 )
available for such cases.

While we could add an additional class to every link, it probably makes more sense to use the mixin so that every future link is also covered.

Additionally, this change also fixes the "wrong" light-blue colour on non-visited links when focussed.

## Visual changes on non-visited link focus

**Before**
Screenshot 2020-03-05 at 13 28 20

**After**
Screenshot 2020-03-05 at 13 29 46

## How to review
- Build locally
- keyboard tab to a visited external link, check text colour (should be black)
- keyboard tab to an internal link, check text colour (should be black)

### DIFF
--- a/src/layouts/govuk.screen.scss
+++ b/src/layouts/govuk.screen.scss
@@ -70,16 +70,6 @@ abbr {
 }
 
 
-a.govuk-link:not([href*="//"]) {
-  &:visited, &:link {
-    color: $govuk-link-colour;
-  }
-
-  &:active {
-    color: $govuk-link-active-colour;
-  }
-
-  &:hover {
-    color: $govuk-link-hover-colour;
-  }
+.govuk-link:not([href*="//"]) {
+  @include govuk-link-style-no-visited-state;
 }


paroxp:
I was not aware of the mixin. That's awesome!

kr8n3r:
that's fair, I wonder if that was to solve a problem, like people not reading or aware that there is documentation, even though they had visited it before)

We can look at unifying the colour (or not) in the future, but for now I've left the visited colour for external links, while still fixing the light blue on yellow

Screenshot 2020-03-06 at 09 13 17

Screenshot 2020-03-06 at 09 13 28

tlwr:
There is more context in this PR: https://github.com/alphagov/paas-admin/pull/532

This change is good, :shipit:

PR #696, opened 2020-03-06
Introduce Marketplace

paroxp:
What
----

Adding new public page, representing services we offer and plans for each service to nicely represent to our users.

How to review
-------------

- Deploy to your environment (or see mine)
- As unauthenticated user, navigate to `/marketplace`
- Attempt to use the service
- Roughly decide if the information is correct


src/components/marketplace/views.tsx
```diff
@@ -0,0 +1,456 @@
...
+
+export function Tab(props: ITabProperties): ReactElement {
+  const classess = ['govuk-tabs__list-item'];
+  if (props.active) {
+    classess.push('govuk-tabs__list-item--selected');
+  }
+
+  return (
+    <li className={classess.join(' ')}>
+      <a href={props.href} className="govuk-link">
```
kr8n3r: can we use the corrected class from this fix? https://github.com/alphagov/paas-admin/pull/692

kr8n3r:
Screenshot 2020-03-06 at 13 51 42

mobile layout is a bit broken, because `figure` has some browser styles
Screenshot 2020-03-06 at 13 58 44

if we add `margin: 0` to `.marketplace-list figure ` they are contained in the view port

-----

Logo on the product page on mobile is a tad big, can we reduce that size?
maybe set a max height on `.service-details figure div img`?

Screenshot 2020-03-06 at 14 07 00

----
Not convinced Tabs component is the best UI for this information - what happens when there are multiple versions available - but this is something we can iterate on post launch

paroxp:
Addressed comments above. By applying the changes, I got concerned about the table outgrowing the mobile view. Typical issue. I still have no clue how to fix properly.

kr8n3r:
@paroxp thanks for making the changes, looks much better now.
do you want to make the change as in https://github.com/alphagov/paas-admin/pull/697/commits/8e7393050e6669dd82df9763cfed08e933ff953e to remove the JS error?

We have a ticket to look at tables on mobile in the upcoming sprint.

while i don't yet full understand all the code code, it looks to work and has tests, so happy to approve

paroxp:
I've removed the `data-module` attribute, thanks!

You can indeed approve, however merging would be sadly blocked as the change is dependant on other PRs from this story. :(

tlwr:
Rebased this on master, paired with Jani, and made some changes:

- "Price: free" or "Price $0.1 per hour" has been removed as a column, if users need this we can add it later. Pricing can be done via the calculator which we could make more prominent
- Added a new column "Available in trial" which uses the `free` boolean flag on the ServicePlan type
- Use the raw plan name rather than the display name, this is to aid users who are trying to work out which plan to use, they can copy and paste the plan name directly, rather than divining that "Tiny" when version=11 is "tiny-unencrypted-11"
- Wrapped the marketplace table in a scrollable container, so that everything works nicely on mobile

Whilst doing the review, I noticed some of the links were wrong, so I'm going to raise a paas-cf pr, get that out, then merge and deploy this

PR #697, opened 2020-03-06
Only initialise javascript for components in use

kr8n3r:
## What
Initialise javascript for component we actually use, not all of them.
Remove the `data-module="govuk-tabs"` attribute from the tab components that are used links to other pages

## Why
At the moment we have a console error for the tabs component as we're only making it look like tabs, but links link to actual pages not content panels and the missing attribute throws an error

## How to review
- build locally, check pages with tabs

### DIFF
--- a/src/frontend/javascript/init.js
+++ b/src/frontend/javascript/init.js
@@ -1,3 +1,29 @@
-(function() {
-  window.GOVUKFrontend.initAll();
-})();
+// there is ever only one header per page
+var $headerMenuButton = document.querySelector('[data-module="govuk-header"]');
+var GOVUKHeader = window.GOVUKFrontend.Header;
+if ($headerMenuButton) {
+  new GOVUKHeader($headerMenuButton).init();
+}
+
+var $buttons = document.querySelectorAll('[data-module="govuk-button"]');
+var GOVUKButton = window.GOVUKFrontend.Button;
+if ($buttons) {
+  for (var i = 0; i < $buttons.length; i++) {
+    new GOVUKButton($button).init();
+  };
+}
+
+var $details = document.querySelectorAll('[data-module="govuk-details"]');
+var GOVUKDetails = window.GOVUKFrontend.Details;
+if ($details) {
+  for (var i = 0; i < $details.length; i++) {
+    new GOVUKDetails($detail).init();
+  };
+}
+
+// there is ever only one error summuary per page
+var $errorSummary = document.querySelector('[data-module="govuk-error-summary"]');
+var GOVUKErrorSummary = window.GOVUKFrontend.ErrorSummary;
+if ($errorSummary) {
+  new GOVUKErrorSummary($errorSummary).init();
+}
--- a/src/components/applications/views.tsx
+++ b/src/components/applications/views.tsx
@@ -58,7 +58,7 @@ export function ApplicationTab(props: IApplicationTabProperties): ReactElement {
         {props.application.entity.name}
       </h1>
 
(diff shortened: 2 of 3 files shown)


richardTowers:
👍  to fixing the tabs error - it's been annoying me too, but I never got around to investigating properly.

Should the tabs component have a no-javascript mode perhaps? Like if there were a class we could put on our tabs to say "hey javascript, it's okay you can ignore these tabs, they don't need enhancement".

The suggested fix (not initialising the tabs JS in our frontend) will only work so long as all our tabs are links, not content panels. We'll need to revisit it if we add any _real_ tabs in the future.

Or perhaps we're just misusing the tabs component entirely?

All that said, I think this is better than what we've got at the moment (errors in production), so I'd vote yes on just merging it :).

src/frontend/javascript/init.js
```diff
@@ -1,3 +1,43 @@
...
+// else good'ol for loop
+var nodeListToArray = function (nodes, callback, scope) {
+  if (window.NodeList && Array.from) {
+    return Array.from(nodes, callback)
+  }
+
+  for (var i = 0; i < nodes.length; i++) {
+    callback.call(scope, nodes[i], i, nodes)
+  }
+}
```
richardTowers: It feels like it's maybe overkill to create this function just to save ourselves two for loops 😄 - it's not even that much clearer:

```js
  nodeListToArray($buttons, function ($button) {
    new GOVUKButton($button).init();
  });

  // Versus
  for (var i = 0; i< $buttons.length, i++ {
    new GOVUKButton($buttons[i]).init();
  });
```

src/frontend/javascript/init.js
```diff
@@ -1,3 +1,43 @@
...
+
+// there is ever only one header per page
+var $headertMenuButton = document.querySelector('[data-module="govuk-header"]');
+var GOVUKHeader = window.GOVUKFrontend.Header;
+if ($headertMenuButton) {
+  new GOVUKHeader($headertMenuButton).init();
```
richardTowers:
```suggestion
var $headerMenuButton = document.querySelector('[data-module="govuk-header"]');
var GOVUKHeader = window.GOVUKFrontend.Header;
if ($headerMenuButton) {
  new GOVUKHeader($headerMenuButton).init();
```

kr8n3r: 🤦‍♂ thanks

kr8n3r:
> 👍 to fixing the tabs error - it's been annoying me too, but I never got around to investigating properly.
> 
> Should the tabs component have a no-javascript mode perhaps? Like if there were a class we could put on our tabs to say "hey javascript, it's okay you can ignore these tabs, they don't need enhancement".
> 
> The suggested fix (not initialising the tabs JS in our frontend) will only work so long as all our tabs are links, not content panels. We'll need to revisit it if we add any _real_ tabs in the future.
> 
> Or perhaps we're just misusing the tabs component entirely?
> 
> All that said, I think this is better than what we've got at the moment (errors in production), so I'd vote yes on just merging it :).

there isn't a no-js mode but as we're setting the active class of page render, so we could just take away the `data-module="govuk-tabs"` to prevent javascript from running.

perhaps it's not the best use of tabs as page links, but we might be ab le to iterate on that.

src/frontend/javascript/init.js
kr8n3r: yes we can do that for simplicity for now.

richardTowers:
👍  - I'd be happy to approve this

paroxp:
Nice. Let's roll!

PR #711, opened 2020-03-10
Improve linter

paroxp:
What
----

I've noticed that some tests are not actually working, due to bugs in code.

We'd like to be more rigorous about the code to prevent certain problems
and help with debugging. Moving away from TSLint has given too much room
for error.

Some plugins come in helpful. For instance the a11y plugin, will help us
remember about accessibility issues when writing or refactoring the
code.

Most of these changes are adding spacing, reordering lines, making sure
the async await has been used properly.

There are however few exceptions...

Some codebase has been simplified and fix not to throw compiler errors.

Some tests have been "straightened" in a way, that proved the tests have
never been running or working properly.

Some codebase have been fixed, as the tests have FINALLY highlighted it
being broken.

How to review
-------------

- Deploy to your environment (or checkout mine)
- Click around and attempt to use crucial functionality (permissions for instance)


src/components/platform-admin/views.tsx
```diff
@@ -95,7 +95,7 @@ function Costs(props: IFormProperties): ReactElement {
         </div>
 
         <div className="govuk-form-group">
-          <label className="govuk-label govuk-date-input__label">Format</label>
+          <label className="govuk-label govuk-date-input__label" htmlFor="view-costs-format">Format</label>
```
kr8n3r: 👏

src/components/reports/views.tsx
```diff
@@ -460,7 +462,7 @@ export function CostByServiceReport(props: ICostByServiceReport): ReactElement {
         <tbody className="govuk-table__body">
           {props.billablesByOrganisationAndService.map(record => (
             <tr key={record.orgGUID} className="govuk-table__row">
-              <td className="govuk-table__cell" scope="row">
```
kr8n3r: this needs to be kept

src/components/reports/views.tsx
```diff
@@ -518,7 +520,7 @@ export function CostByServiceReport(props: ICostByServiceReport): ReactElement {
         <tbody className="govuk-table__body">
           {props.billablesByOrganisationAndSpaceAndService.map(record => (
             <tr key={record.spaceGUID} className="govuk-table__row">
-              <td className="govuk-table__cell" scope="row">
```
kr8n3r: keep

src/components/spaces/views.tsx
```diff
@@ -277,7 +274,7 @@ export function SpacesPage(props: ISpacesPageProperties): ReactElement {
         {props.organization.entity.name}.
       </p>
 
-      <details className="govuk-details" role="group">
```
kr8n3r: keep https://github.com/alphagov/govuk-frontend/blob/b57ade99ac353a944620d818c1f89e1005935181/src/govuk/components/date-input/template.njk#L80

src/layouts/template.tsx
```diff
@@ -139,9 +140,10 @@ export class Template {
   private EnableClientSideJavaScript(): ReactElement {
     return (
       <script
+        // eslint-disable-next-line react/no-danger
         dangerouslySetInnerHTML={{
           __html:
-            `document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');`,
+            'document.body.className = ((document.body.className) ? document.body.className + \' js-enabled\' : \'js-enabled\');',
```
kr8n3r: hmmm, we should check that classes are correct when this is built

paroxp: I agree. I seem to recall this may be problematic. This is the linter auto-fixing things.

paroxp: Tested in my environment and the `js-enabled` has been correctly applied 🎉

src/components/spaces/views.tsx
paroxp: Reverted back in separate commit, with an exception to the linter.

PR #720, opened 2020-03-16
Remove any spaces from an email address

paroxp:
What
----

It appears that some people copy and paste an email address with a
trailing space or miss type it in the first place.

Whilst our validation sees that as correct email, UAA API would respond
with an error, which we do not handle...

We have the principle of making users pass the first time, therefore
we're stripping out the space instead of presenting an error to them
complaining about the space.

How to review
-------------

- Deploy to your dev env (or see rafalp)
- Attempt to invite new user, whilst appending their email address with a space
- Expect no error page to appear

### DIFF
--- a/src/components/org-users/controllers.tsx
+++ b/src/components/org-users/controllers.tsx
@@ -219,9 +219,7 @@ function validatePermissions({
   return errors;
 }
 
-function validateEmail({
-  email,
-}: IUserPostBody): ReadonlyArray<IValidationError> {
+function validateEmail({ email }: IUserPostBody): ReadonlyArray<IValidationError> {
   const errors: Array<IValidationError> = [];
 
   if (!email || !VALID_EMAIL.test(email)) {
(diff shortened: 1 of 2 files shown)


src/components/org-users/controllers.tsx
```diff
@@ -472,17 +470,17 @@ export async function inviteUser(
       throw new ValidationError(errors);
     }
 
-    const uaaUser = await uaa.findUser(userBody.email!);
+    const email = userBody.email!.replace(' ', '');
```
tlwr:
```suggestion
    const email = userBody.email!.replace(/\s/g, '');
```

Replace all whitespace, not just spaces

Replace whitespace globally, not just the first instance (so that both leading and trailing whitespace is handled)

paroxp: Good shout.

pauldougan:
great, thanks for fixing this so quicly

PR #787, opened 2020-04-08
Implement Service Logs

paroxp:
What
----

We'd like to allow RDS provided services, to obtain logs for the
instances in question.

Majority of the work is done for us by the AWS API and SDK.

How to review
-------------

- Code review
- Add IAM policies with accordance to alphagov/paas-aws-account-wide-terraform#214
- Deploy to your devenv
- Navigate to `admin` > `billing` > Backing Services > `billing-db` > Logs
- Use the feature

Preview
--------

image

### DIFF
--- a/src/components/app/router.ts
+++ b/src/components/app/router.ts
@@ -75,8 +75,7 @@ export const router = new Router([
   {
     action: services.viewService,
     name: 'admin.organizations.spaces.services.view',
-    path:
-      '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID',
+    path: '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID',
   },
   {
     action: serviceMetrics.viewServiceMetrics,
--- a/src/components/app/router.ts
+++ b/src/components/app/router.ts
@@ -108,6 +108,16 @@ export const router = new Router([
       '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/events/:eventGUID',
   },
   {
+    action: services.listServiceLogs,
+    name: 'admin.organizations.spaces.services.logs.view',
+    path: '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs',
+  },
+  {
+    action: services.downloadServiceLogs,
+    name: 'admin.organizations.spaces.services.logs.download',
+    path: '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs/download',
+  },
+  {
     action: orgUsers.listUsers,
     name: 'admin.organizations.users',
     path: '/organisations/:organizationGUID/users',
--- a/src/components/services/controllers.mocked.test.tsx
+++ b/src/components/services/controllers.mocked.test.tsx
@@ -1,0 +1,177 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+import { RDS } from 'aws-sdk';
+
+import CloudFoundryClient from '../../lib/cf';
+import { createTestContext } from '../app/app.test-helpers';
+import { IContext } from '../app/context';
+
+import { downloadServiceLogs, listServiceLogs } from './controllers';
+
+jest.mock('aws-sdk');
+jest.mock('../../lib/cf');
+
+const mockCustomService = { metadata: { guid: 'CUSTOM_SERVICE' } };
+const mockServiceInstance = { entity: { name: 'mydb' }, metadata: { guid: 'SERVICE_INSTANCE_GUID' } };
+const mockServiceRedis = { entity: { label: 'redis' } };
+const mockServicePostgres = { entity: { label: 'postgres' } };
+const mockOrganization = { entity: { name: 'org-name' }, metadata: { guid: 'ORG_GUID' } };
+const mockSpace = { entity: { name: 'space-name' }, metadata: { guid: 'SPACE_GUID' } };
+
+const ctx: IContext = createTestContext();
+const CFClient = CloudFoundryClient as jest.Mock;
+
+describe(listServiceLogs, () => {
+  beforeEach(() => {
+    CFClient.mockClear();
+    // @ts-ignore
+    RDS.mockClear();
+  });
+
+
+  it('should throw an error if service is custom user service', async () => {
+    CFClient.prototype.userServices.mockReturnValueOnce(Promise.resolve([ mockCustomService ]));
+
+    await expect(
+      listServiceLogs(ctx, { organizationGUID: 'ORG_GUID', serviceGUID: 'CUSTOM_SERVICE', spaceGUID: 'SPACE_GUID' }),
+    ).rejects.toThrowError(/Service Logs are only available for Postgres and MySQL instances/);
+  });
+
+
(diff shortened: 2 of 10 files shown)


src/components/services/services.scss
```diff
@@ -0,0 +1,9 @@
+.service-log-list {
```
kr8n3r, line 1: can we do 
```.service-log-list-item__attribute {
    &:last-child {
      display: block;
    }
}
```
instead without the nesting?

kr8n3r:
@paroxp could you add a screenshot if pos?

paroxp:
Done @kr8n3r

kr8n3r:
thanks, that looks fine.
in the demo i used `aria-describedBy="unique-id"` on the link and then ` id="unique-id"` on the paragraph so that on focus all that info below was read by the screenreader.

I can't see this in the code, unless I'm missing something

kr8n3r:
also are we able to reverse-sort them, latest at the top?

paroxp:
Done and done.

kr8n3r:
magic. 👍 for FE, pester engineers for other code reviews

tlwr:
In terms of code, this can go out

Could we improve the user experience somewhat by offering the files in a more structured format, e.g. a table

tlwr:
Rafal aware'd me on the reason why this page looks how it does, from a Slack thread with Jani:

> navigation a table requires more effort for the users with assistive tech. by providing a simple list and using rer describedby we are giving them all the information they need and the announcement when they reach the list says x number of items

Just gonna double check my code review then we can :shipit:

src/components/services/controllers.mocked.test.tsx
```diff
@@ -0,0 +1,198 @@
...
+        promise: async () => await Promise.resolve({}),
+      }),
+    });
+
+    const response = await listServiceLogs(ctx, {
+      organizationGUID: 'ORG_GUID', serviceGUID: 'SERVICE_INSTANCE_GUID', spaceGUID: 'SPACE_GUID',
+    });
+
+    expect(response.body).toContain('mydb - Service Metrics Logs');
+    expect(response.body).toContain('There are no log files available at this time. Please comeback later.');
```
tlwr, line 95:
```suggestion
    expect(response.body).toContain('There are no log files available at this time. Please come back later.');
```

Typo, and I'm not sure we need the user to come back later

src/components/services/controllers.tsx
```diff
@@ -72,3 +76,115 @@ export async function viewService(
...
+
+  const rds = new RDS();
+  const fileList = await rds.describeDBLogFiles({
+    DBInstanceIdentifier: `rdsbroker-${serviceInstance.metadata.guid}`,
+    MaxRecords: 72,
+  }).promise();
+
+  const template = new Template(
+    ctx.viewContext,
+    `${serviceInstance.entity.name} - Service Metrics Logs`,
```
tlwr, line 116:
```suggestion
    `${serviceInstance.entity.name} - Service Logs`,
```

src/components/services/controllers.tsx
```diff
@@ -72,3 +76,115 @@ export async function viewService(
...
+
+  const serviceInstance = await cf.serviceInstance(params.serviceGUID);
+  const service = await cf.service(serviceInstance.entity.service_guid);
+
+  if (!['postgres', 'mysql'].includes(service.entity.label)) {
+    throw new UserFriendlyError(UNSUPPORTED_SERVICE_LOGS_REQUEST);
+  }
+
+  const rds = new RDS();
+  const fileList = await rds.describeDBLogFiles({
```
tlwr, line 109: I'm assuming AWS returns these in the correct order, so we don't have to paginate

src/components/services/views.tsx
```diff
@@ -166,3 +202,60 @@ export function ServicePage(props: IServicePageProperties): ReactElement {
...
+                    filename: file.LogFileName,
+                    organizationGUID: props.organizationGUID,
+                    serviceGUID: props.service.metadata.guid,
+                    spaceGUID: props.spaceGUID,
+                  })}
+                  name={file.LogFileName}
+                  size={file.Size}
+                />)}
+              </ul>
+            : <p className="govuk-body">There are no log files available at this time. Please comeback later.</p>
```
tlwr, line 255:
```suggestion
            : <p className="govuk-body">There are no log files available at this time. Please come back later.</p>
```

Comment above about wording

src/components/app/router.ts
```diff
@@ -108,6 +107,16 @@ export const router = new Router([
...
   },
+  {
+    action: services.listServiceLogs,
+    name: 'admin.organizations.spaces.services.logs.view',
+    path: '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs',
+  },
+  {
+    action: services.downloadServiceLogs,
+    name: 'admin.organizations.spaces.services.logs.download',
+    path: '/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs/download',
```
tlwr, line 118: This could be:

```
/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs/:logFileName/download
```

which would allow us to build:

```
/organisations/:organizationGUID/spaces/:spaceGUID/services/:serviceGUID/logs/:logFileName/view
```

in the future. We can do this in a separate PR if we ever build view functionality.

src/components/services/controllers.tsx
```diff
@@ -72,3 +76,115 @@ export async function viewService(
...
+        spaceGUID: space.metadata.guid,
+      }),
+      text: serviceInstance.entity.name,
+    },
+    { text: 'Logs' },
+  ]);
+
+  return {
+    body: template.render(<ServiceLogsPage
+      files={(fileList.DescribeDBLogFiles || []).reverse() as ReadonlyArray<IServiceLogItem>}
```
tlwr, line 139:
```suggestion
      files={(fileList.DescribeDBLogFiles || []) as ReadonlyArray<IServiceLogItem>}
```

Personal preference, but the most recent log file first makes more sense to me

src/components/app/router.ts
paroxp: I'd agree. The problem is that the file names are: `errors/postgres.log-[timestamp]` which will likely mess up the url. Sadly that's the only "selectable" field allowed by AWS.

src/components/services/controllers.tsx
paroxp: `.reverse()` gives us the recent to the top :) See Jani's comment above.

src/components/services/views.tsx
paroxp: Deleted.

src/components/services/controllers.tsx
paroxp: It would appear that AWS returns all the files (162 at once) in a asc order. Hence the following `.reverse()`.

src/components/services/controllers.tsx
paroxp: Done.

src/components/services/controllers.mocked.test.tsx
paroxp: Deleted.

tlwr:
Big fan, gonna merge

Issue #103, opened 2018-10-10
A user with global auditing can reach the "user details" page but it would fail


keymon:
# What?

A user with Global Auditor role in a CF installation can go to an org, see all the org users, and then click on any of these users. But when displaying the user page, it fails.

# Expected

The auditor should not be able to "click" on the user details if it has no permissions, or a more informative error should be displayed.

46bit:
Should be fixed by #882. Global Auditors still can't see the user page and still get linked to it, but they get a useful error message.

PR #882, opened 2020-05-20
Be nice to Global Auditors accessing user pages

46bit:
What
----

In #103 we recorded that Global Auditors can't see the pages corresponding to individual users. This is part of Cloud Foundry's permission model and not something we really want to change. But until now Global Auditors have got an unfriendly, unclear error message.

This commit changes the page for individual users so that Global Auditors will get a new informative message telling them what to do:

> your "global auditor" permissions do not allow viewing users. sorry. please ask someone who will have more permissions, such as an engineer.

A better fix would be to refine the concept of what a global admin is but that would be a larger project.

How to review
-------------

* See if the tests go green?
* Check you're OK with this simple-as-possible implementation.
* Double-check this actually works outside of unit tests.

Who can review
---------------

Not @46bit.

### DIFF
--- a/src/components/users/controllers.test.tsx
+++ b/src/components/users/controllers.test.tsx
@@ -1,7 +1,6 @@
 import jwt from 'jsonwebtoken';
 import nock from 'nock';
 
-import { spacesMissingAroundInlineElements } from '../../layouts/react-spacing.test';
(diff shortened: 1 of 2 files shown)


46bit:
I wrote some code! ^_^

Unfortunately this still took a lot out of me. :(

PR #893, opened 2020-05-27
Add password reset functionality

paroxp:
What
----

We'll be sending out password reset email to our tenants. Notify
requires us to point out a template to be used for email sending
functionality.

We have a wrapper around notify's SDK, which is limited to our needs.
We're adding a new function that will specifically target the password
reset email.

We're able to use UAA headlessly via the API to perform some actions.
These include the password reset flow. We're simply adding functions
calling for these specific endpoints, with some additional data.

It seems that using both JSON.stringify and axios, messes up with the
attribute enough, for the UAA API to return bad request response.

This deserves testing the user invitation functionality...

As a bonus, I've had a look into the navigation component.

We're adding more and more pages which do not require authentication.
This makes it unreasonable, to ignore the navigation always say things
like `Sign out`. Instead, we're adding a flag, which will trigger
different menu items depending on the state of the session.

How to review
-------------

- Deploy alphagov/paas-cf#2328 to your dev env - or manually setup your env
- Deploy pazmin from this branch to your environment
- Attempt to run through entire forgotten password flow
- Check wording please - I made it all up...

⚠️ This needs to be merged before the alphagov/paas-cf#2328 ⚠️ 
-------------------------------------------------------------------

### DIFF
--- a/src/components/users/controllers.tsx
+++ b/src/components/users/controllers.tsx
@@ -193,7 +193,7 @@ export async function resetPasswordObtainToken(
   const resetCode = await uaa.obtainPasswordResetCode(uaaUser.userName);
 
   const url = new URL(ctx.app.domainName);
-  url.pathname = '/password/reset';
+  url.pathname = '/password/confirm-reset';
   url.searchParams.set('code', resetCode);
 
   await notify.sendPasswordReminder(email, url.toString());
--- a/src/components/users/controllers.tsx
+++ b/src/components/users/controllers.tsx
@@ -234,7 +234,7 @@ export async function resetPassword(ctx: IContext, _params: IParameters, body: I
       body: template.render(<PasswordResetSetPasswordForm
         csrf={ctx.viewContext.csrf}
         code={body.code}
-        passwordMissmatch={true}
+        passwordMismatch={true}
       />),
       status: 400,
     };
--- a/src/components/users/controllers.tsx
+++ b/src/components/users/controllers.tsx
@@ -1,3 +1,4 @@
+import { capitalize } from 'lodash';
 import React from 'react';
 
 import { Template } from '../../layouts';
@@ -115,7 +116,7 @@ export async function resetPasswordObtainToken(
   ctx: IContext, params: IParameters, body: IPasswordResetBody,
 ): Promise<IResponse> {
   const VALID_EMAIL = /[^.]@[^.]/;
-  const email = body.email;
+  const email = (/* istanbul ignore next */ body.email || '').toLowerCase();
   const template = new Template(ctx.viewContext, 'Request password reset');
 
   if (!email || !VALID_EMAIL.test(email)) {
@@ -135,17 +136,51 @@ export async function resetPasswordObtainToken(
     },
   });
 
-  template.title = 'Password reset successfully requested';
-  const successPage = <PasswordResetSuccess
-    title="Password reset successfully requested"
-    message="You should receive an email shortly with further instructions."
-    footnote="If you do not receive an email in the next 24 hours, your account may not exist in the system or you could have setup authentication with Google or Microsoft Single Sign-on."
-  />;
+  const accounts = new AccountsClient({
+    apiEndpoint: ctx.app.accountsAPI,
+    logger: ctx.app.logger,
+    secret: ctx.app.accountsSecret,
+  });
 
-  const user = await uaa.findUser(email);
-  if (!user || user.origin !== 'uaa') {
+  const accountsUser = await accounts.getUserByEmail(email);
+  if (!accountsUser) {
+    // 100ms delay to stop people iterating through users
+    await new Promise(finish => setTimeout(finish, 100));
+
+    template.title = 'Error: User not found';
+
     return {
-      body: template.render(successPage),
+      body: template.render(<PasswordResetRequest csrf={ctx.viewContext.csrf} invalidEmail={true} />),
+      status: 404,
+    };
+  }
+
+  const uaaUser = await uaa.getUser(accountsUser.uuid)
+  if (!uaaUser) {
+    // 100ms delay to stop people iterating through users
+    await new Promise(finish => setTimeout(finish, 100));
+
+    template.title = 'Error: User not found';
+
+    return {
+      body: template.render(<PasswordResetRequest csrf={ctx.viewContext.csrf} invalidEmail={true} />),
+      status: 404,
+    };
+  }
+
+  if (uaaUser.origin != 'uaa') {
+    // 100ms delay to stop people iterating through users
+    await new Promise(finish => setTimeout(finish, 100));
+
+    const idpNice = capitalize(uaaUser.origin);
+    template.title = `Error: You have enabled ${idpNice} single sign-on`;
+
+    return {
+      body: template.render(
+        <PasswordResetRequest
+          csrf={ctx.viewContext.csrf}
+          invalidEmail={true} />,
+      ),
     };
   }
 
@@ -155,14 +190,19 @@ export async function resetPasswordObtainToken(
       passwordReset: ctx.app.notifyPasswordResetTemplateID,
     },
   });
-  // POTENTIAL: Wrap the next line in try catch, and if UAA returns 409, send user email saying they're using SSO.
-  const code = await uaa.obtainPasswordResetCode(user.userName);
+  const resetCode = await uaa.obtainPasswordResetCode(uaaUser.userName);
 
   const url = new URL(ctx.app.domainName);
   url.pathname = '/password/reset';
-  url.searchParams.set('code', code);
+  url.searchParams.set('code', resetCode);
 
   await notify.sendPasswordReminder(email, url.toString());
+
+  template.title = 'Password reset successfully requested';
+  const successPage = <PasswordResetSuccess
+    title="Password reset successfully requested"
+    message="You should receive an email shortly with further instructions."
+  />;
 
   return {
     body: template.render(successPage),
--- a/src/components/users/controllers.tsx
+++ b/src/components/users/controllers.tsx
@@ -220,6 +220,45 @@ export async function resetPasswordProvideNew(ctx: IContext, params: IParameters
   });
 }
 
+export function checkPasswordAgainstPolicy(pw: string): { valid: boolean; message?: string } {
+  if (pw.length < 12) {
+    return {
+      valid: false,
+      message: 'Your password should be 12 characters or more',
+    }
+  }
+
+  if (!pw.match(/[a-z]/)) {
+    return {
+      valid: false,
+      message: 'Your password should contain a lowercase character',
+    }
+  }
+
+  if (!pw.match(/[A-Z]/)) {
+    return {
+      valid: false,
+      message: 'Your password should contain an uppercase character',
+    }
+  }
+
+  if (!pw.match(/[0-9]/)) {
+    return {
+      valid: false,
+      message: 'Your password should contain a number',
+    }
+  }
+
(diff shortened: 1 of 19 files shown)


src/components/users/views.tsx
```diff
@@ -101,3 +121,168 @@ export function UserPage(props: IUserPageProps): ReactElement {
...
+    <div className="govuk-grid-column-two-thirds">
+      <h1 className="govuk-heading-l">Password reset</h1>
+      <p className="govuk-body">
+        We will send you an activation link via email. If you do not receive an
+        email in the next 24 hours, your account may not exist in the system or
+        you could have setup authentication with Google or Microsoft Single
+        Sign-on.
+      </p>
+
+      {props.invalidEmail ? <div
```
kr8n3r, line 136: can we move the error-summary to the top of the page as documented in https://design-system.service.gov.uk/components/error-summary/

src/components/users/views.tsx
```diff
@@ -101,3 +121,168 @@ export function UserPage(props: IUserPageProps): ReactElement {
...
+        className="govuk-error-summary"
+        aria-labelledby="error-summary-title"
+        role="alert"
+        tabIndex={-1}
+        data-module="govuk-error-summary"
+      >
+        <h2 className="govuk-error-summary__title" id="error-summary-title">
+          Error validating the request
```
kr8n3r, line 144: ‘There is a problem’ in the summary title

src/components/users/views.tsx
```diff
@@ -101,3 +121,168 @@ export function UserPage(props: IUserPageProps): ReactElement {
...
+        <input type="hidden" name="_csrf" value={props.csrf} />
+
+        <div className={`govuk-form-group ${props.invalidEmail ? 'govuk-form-group--error' : ''}`}>
+          <label className="govuk-label" htmlFor="email">
+            Email address
+          </label>
+
+          {props.invalidEmail ? <span id="email-error" className="govuk-error-message">
+            <span className="govuk-visually-hidden">Error:</span>{' '}
+            Valid email address must be provided
```
kr8n3r, line 166: can we update the error message, something like https://design-system.service.gov.uk/patterns/email-addresses/#error-messages

paroxp:
I've addressed all your comments.

I have to say, keeping titles and H1s in sync will be a pain... We should think of a pattern for that 🤔

src/components/users/controllers.tsx
```diff
@@ -89,3 +96,119 @@ export async function getUser(
...
+  });
+
+  template.title = 'Password reset successfully requested';
+  const successPage = <PasswordResetSuccess
+    title="Password reset successfully requested"
+    message="You should receive an email shortly with further instructions."
+    footnote="If you do not receive an email in the next 24 hours, your account may not exist in the system or you could have setup authentication with Google or Microsoft Single Sign-on."
+  />;
+
+  const user = await uaa.findUser(email);
```
tlwr, line 139: Would it be worth making a call to the accounts API instead of UAA and explicitly returning a response:

```
const accountsUser = accountsClient.getUserByEmail(email)
if (!user) return "not found"

const uaaUser = uaaClient.getUserByGUID(user.guid)
if (uaaUser.origin != 'uaa') return "use single sign on"

const code = await uaa.obtainPasswordResetCode(uaaUser.userName);
...
```

My understanding is

- If the user is not using UAA to manage their account, UAA may not know their email
- PaaS accounts is the canonical store of email addresses
- If they are using SSO we should just tell them they are using SSO (as per the "It's done when" on the story)
- If we are worried about security we can introduce an artificial delay on this page to reduce the rate at which someone can enumerate email addresses.`

LeePorte: Using the Accounts API is a sensible apporach here. Just wondering if there are any gotchas that we've not throught of that Rafal already has.

tlwr: The "It's done when" is:

> a SSO enabled tenant receives a helpful prompt on the page in response to asking for a password reset rather than nothing.

The current implementation of this PR is:

- try to find the user based on their email address in UAA
- if we cannot find the user or the user is using SSO, give them a page which says we have sent an email (which we haven't sent) and to wait 24h
- if we can find the user, and the user is not using SSO, send them an email

I think there are some improvements which block this from being merged:
- We should not say we have sent an email when we haven't sent an email
- If we cannot find the user we should say "your user was not found"
- If the user is using SSO we should say "you are using SSO"
- If we have security concerns we can artificially introduce a time delay in rendering the response, or we can send an email instead

tlwr:
Rebased onto master and now using PaaS accounts and have added user friendly error messages

Screenshot showing error message when using has SSO enabled using Google

Screenshot showing User not found error message

Screenshot showing success state

I'm now adding a password policy, validator, and good guidance

kr8n3r:
some DS guidance https://design-system.service.gov.uk/patterns/passwords#helping-users-choose-strong-passwords

tlwr:
Screenshot showing confirmation when not meeting the password policy for length

Schreenshot showing confirmation when not meeting the password policy for uppercase

src/components/users/views.tsx
```diff
@@ -101,3 +126,195 @@ export function UserPage(props: IUserPageProps): ReactElement {
...
+          {props.passwordMismatch ? <span id="password-error" className="govuk-error-message">
+            <span className="govuk-visually-hidden">Error:</span>{' '}
+            You need to type in the same password twice
+            </span> : <></>}
+
+          {props.passwordDoesNotMeetPolicy ? <span id="password-error" className="govuk-error-message">
+            <span className="govuk-visually-hidden">Error:</span>{' '}
+            {props.passwordDoesNotMeetPolicyMessage!}
+          </span> : <></>}
+
```
kr8n3r, line 283: each input should be in their own `govuk-form-group`, currently they're both in the same one, hence no spacing
and error message styling (as `form-group--error` class needs to be added to the div.

when there's an error on the input it should have the below styling
Screenshot 2020-06-12 at 12 55 35

tlwr:
It is better now

---
image
---
image

kr8n3r:
:ship from FE side

src/components/users/controllers.tsx
```diff
@@ -95,3 +106,208 @@ export async function getUser(
...
+  }
+
+  const template = new Template(ctx.viewContext, 'Password reset');
+
+  return await Promise.resolve({
+    body: template.render(<PasswordResetSetPasswordForm csrf={ctx.viewContext.csrf} code={params.code} />),
+  });
+}
+
+export function checkPasswordAgainstPolicy(pw: string): { valid: boolean; message?: string } {
```
paroxp, line 223:
```suggestion
export function checkPasswordAgainstPolicy(pw: string): { valid: boolean; messages?: ReadonlyArray<string> } {
```

That way, user won't have to re-submit the form multiple times before they get their password right.

src/components/users/controllers.tsx
```diff
@@ -95,3 +106,208 @@ export async function getUser(
...
+  }
+
+  if (!pw.match(/[-_+:;<>[\]()#@£%^&!/]/)) {
+    return {
+      valid: false,
+      message: 'Your password should contain a special character',
+    }
+  }
+
+  return { valid: true };
```
paroxp, line 259: I'm not a biggest fan of this function. It's independent from UAA, meaning we may be passing on one front and failing on another (at some point in time).

It could also lead to frustrated users, just adding random characters without memorising the password causing another need for password reset.

Just saying, no need to act :)

src/components/users/controllers.tsx
```diff
@@ -95,3 +106,208 @@ export async function getUser(
...
+    logger: ctx.app.logger,
+    secret: ctx.app.accountsSecret,
+  });
+
+  const accountsUser = await accounts.getUserByEmail(email);
+  if (!accountsUser) {
+    // 100ms delay to stop people iterating through users
+    await new Promise(finish => setTimeout(finish, 100));
+
+    template.title = 'Error: User not found';
```
paroxp, line 153: I'm not a fan of this error being presented. This has a potential for an attacker to recognise and perform targeted attacks. Delay is good for mass snooping through users, but not targeted.

Just saying, no need to act :)

src/components/users/controllers.tsx
```diff
@@ -95,3 +106,208 @@ export async function getUser(
...
+
+  const uaa = new UAAClient({
+    apiEndpoint: ctx.app.uaaAPI,
+    clientCredentials: {
+      clientID: ctx.app.oauthClientID,
+      clientSecret: ctx.app.oauthClientSecret,
+    },
+  });
+
+  const accounts = new AccountsClient({
```
paroxp, line 142: Only reason that wasn't here since beginning, is that there is a potential for people using CLI not the admin UI to perform certain actions, such as user creation. Meaning our own collection, could get out of date.

I'm happy for it to stay, but it's weird we have two DBs with the same data :trollface:

tlwr: UAA doesn't always store a user's email address

PaaS accounts does

Only operators have permissions to create users without going via Pazmin

src/components/users/controllers.tsx
tlwr: The password policy is more restrictive than what we have in UAA

I think we should make UAA more restrictive to match this password policy

src/components/users/controllers.tsx
tlwr: @kr8n3r is it cool to display all the messages at the same time? if so I'll introduce this change

src/components/users/controllers.tsx
tlwr: A compromise here would be

If the user is not found, sleep a long time (eg 500ms)

If the user is found, but using SSO, send them an email saying they are using SSO?

src/components/users/controllers.tsx
paroxp: I'm not saying it isn't. I mean it in a way, that if there is a weird bug or misconfiguration of UAA to require, say `At least three digits in a password`, our filter could pass it through whilst UAA would reject it.

I understand there is little to no chance of that happening, but if it would, it'd be a pain to investigate for someone who wasn't involved in this review 🤔 

Once again, ignore me - I'm happy with it if you are :)

src/components/users/controllers.tsx
paroxp: You're right, but we as operators are also eligible to make a mistake :)

tlwr: I tripped and fell and ran `curl https://uaa.london.cloud.service.gov.uk/Users -X POST -d '{...}'` 😈

src/components/users/controllers.tsx
paroxp: Yeah, previously there was a comment here with a potential to send out an educational email. It wasn't implemented though, as I was hoping the next screen would be informative enough for the users to "remember" they're using SSO.

src/components/users/controllers.tsx
paroxp: I'm glad you find it funny :D I happen to have done that once after my rejoin xD guilty

paroxp: more precisely, `uaa create-user` :P

src/components/users/controllers.tsx
tlwr: You raise a good point, we aren't returning the UAA error to the user if UAA rejects the `/password_change` POST, although we will return an error to the user because of how the UAA client works.

I think it is a sufficiently rare case that I'm fine with it

src/components/users/controllers.tsx
tlwr: I am going to make the delay more aggressive - 400ms - in lieu of sending an email

PR #949, opened 2020-07-13
return "metrics unavailable" page for unknown svcs

chrisfarms:
the service metrics tab would result in a 500 error if it encountered a
service type that it didn't understand.

there are several ways this could occur:

* A new backing service without metrics implemented yet
* A custom service from a space-scoped broker
* A user provided service

The user-provided-service path was hard coded to display a nicer error
page, but anything else would result in a generic error.

This changes it so that ANY unknown service gets the "metrics
unavailable" page so the user doesn't loose their flow and the tabbed
interface operates as expected.

### DIFF
--- a/src/components/service-metrics/controllers.test.tsx
+++ b/src/components/service-metrics/controllers.test.tsx
@@ -319,7 +319,7 @@ describe('service metrics test suite', () => {
     ).toHaveLength(0);
   });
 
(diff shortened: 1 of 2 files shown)


tlwr:
tests ran locally ( because sadpanda github unreliability)

PR #980, opened 2020-08-05
Only show Logs tab on supported services

paroxp:
What
----

At the moment, all of the services get the logs tab. If the service
doesn't support logs, the tab will simply redirect into a 404 page.

This isn't great user experience.

What we're doing here, is essentially checking if the service is one of
the supported ones, and simply attempt to not display the tab for the
user if it's not supported.

How to review
-------------

- Deploy to your devenv
- Attempt to retrieve logs for postgres/mysql service and any other service
- Expect no 404 behaviours


kr8n3r:
good stuff! if only i had a devenv to push it to

LeePorte:
I'm not seeing a logs tab, where there are no logs any more

PR #990, opened 2020-08-12
Introduce quota management

paroxp:
What
----

Personally I find it difficult to figure out which region does the
organisation belong to, by first logging in to possibly two pazmins, and
then the CLI, figure out what does the quota offer and set it for the
user before informing them of what has been done to their org.

This should reduce the amount of actions I need to perform each time and
help me move on with my life without distractions.

Probably useless to the rest of the team.

How to review
-------------

- Deploy to your environment (or see tlwr)
- Navigate to an organisation
- As an operator find the `Manage quota for this organization`
- Attempt to change the quota

### DIFF
--- a/src/components/app/router.ts
+++ b/src/components/app/router.ts
@@ -36,6 +36,17 @@ export const router = new Router([
     path: '/organisations/:organizationGUID',
   },
   {
+    action: organizations.editOrgQuota,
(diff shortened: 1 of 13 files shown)


tlwr:
Deployed to my dev env, LGTM gonna :shipit:

PR #1013, opened 2020-08-18
Show error when invite additional users is selected but no email address entered

kr8n3r:
What
----
Show error when invite additional users is selected but no email address entered
Users were able to select yes to invite additional users but leave the email fields empty.

Fixes: https://www.pivotaltracker.com/n/projects/1275640/stories/174314885

How to review
-------------

- go to support form, `/support/sign-up`
- select yes to "Invite additional users" but leave their email address fields empty
- check if you see an error message
- play around with other combinations, selecting no, etc - you should see other validation messages but not this one

Who can review
---------------

not @kr8n3r 

0 0 0 0_3000_support_sign-up

### DIFF
--- a/src/components/support/controllers.tsx
+++ b/src/components/support/controllers.tsx
@@ -309,6 +309,19 @@ function validateInviteUsers({ invite_users }: ISignupForm): ReadonlyArray<IVali
   return errors;
 }
 
+function validateAdditionalUserYesButEmpty({ invite_users, additional_users }: ISignupForm): ReadonlyArray<IValidationError> {
+  const errors = [];
+
+  if (invite_users === 'yes' && additional_users?.every(user => user.email === '')) {
+    errors.push({
+        field: 'additional_users-0',
+        message: 'Enter at least one additional user email address',
+      });
+  }
+
+  return errors;
+}
+
 function validateAdditionalUserEmail({ additional_users }: ISignupForm): ReadonlyArray<IValidationError> {
   const errors = [];
   if (additional_users) {
@@ -742,6 +755,7 @@ export async function HandleSignupFormPost(
     ...validateServiceTeam(body),
     ...validatePersonIsManager(body),
     ...validateInviteUsers(body),
+    ...validateAdditionalUserYesButEmpty(body),
     ...validateAdditionalUserEmail(body),
   );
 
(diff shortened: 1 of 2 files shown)


src/components/support/controllers.tsx
```diff
@@ -309,6 +309,28 @@ function validateInviteUsers({ invite_users }: ISignupForm): ReadonlyArray<IVali
   return errors;
 }
 
+function validateAdditionalUserYesButEmpty({ invite_users, additional_users }: ISignupForm): ReadonlyArray<IValidationError> {
+  const errors = [];
+  let userFieldsEmpty = true;
+  if (invite_users === 'yes' && additional_users) {
+    /* eslint-disable @typescript-eslint/prefer-for-of */
+    for (let i = 0; i < additional_users.length; i++) {
```
paroxp, line 317: Any reason to disable?

src/components/support/controllers.tsx
```diff
@@ -309,6 +309,28 @@ function validateInviteUsers({ invite_users }: ISignupForm): ReadonlyArray<IVali
...
+    if (userFieldsEmpty) {
+      errors.push({
+        field: 'additional_users-0',
+        message: 'Enter at least one additional user email address',
+      });
+    }
+  }
+
+  return errors;
+}
```
paroxp, line 332: I think this function could be somewhat simplified.

```typescript
function validateAdditionalUserYesButEmpty({ invite_users, additional_users }: ISignupForm): ReadonlyArray<IValidationError> {
  const errors = [];

  if (additional_users.some(user => user.email !== undefined)) {
    errors.push({
        field: 'additional_users-0',
        message: 'Enter at least one additional user email address',
      });
  }

  return errors;
}
```

What do you think?

paroxp (commented):
Just left some comments around not being sure about the linter being disabled :D

src/components/support/controllers.tsx
kr8n3r: 🏅 forgot about `.some`

kr8n3r: hmm, using some will still return this message when something is entered in field 2 or 3

paroxp (approved):
Nice corrections :)

PR #1016, opened 2020-08-19
Update Org/Space/App/Service/Event titles & headings

kr8n3r:
What
----

Problem:

When users select a "tab" such as ‘Backing services’ the page refreshes and users’ focus is returned to the top of the page. 

As the page title and main heading do not update to reflect the updated page content, some users may be unaware that the page content has changed.

While the [rest of the issue](https://www.pivotaltracker.com/n/projects/1275640/stories/174313243) talks about tabs markup not functioning like tabs in terms of keyboard navigation - they're not tabs, just look like one - for now we'll hold off from redesigning that navigational element, and have auditors test again.

This PR updates the Org/Space/App/Service/Event page titles & headings so they're unique and match when navigating pages.

We've moved the words "Organisation", "Space", "Application", "Service" from the visual text to visually-hidden text that's still part of the heading for screenreader users.

In its place we moved the Org/Space/App/Service name and below added the page title, like "Overview", "Metrics", "Applications" which matches the tab selection made and content on the page.

Visual changes
---

### Before
Screenshot 2020-08-20 at 08 07 41
Screenshot 2020-08-20 at 08 07 33

### After 

Screenshot 2020-08-20 at 08 03 15
Screenshot 2020-08-20 at 08 04 22

**heading also contains visually hidden text "Space" and "Application"**

Screenshot 2020-08-20 at 08 30 49

How to review
-------------

Describe the steps required to test the changes.

Who can review
---------------

### DIFF
--- a/src/components/application-events/controllers.test.tsx
+++ b/src/components/application-events/controllers.test.tsx
@@ -60,7 +60,7 @@ describe('application event', () => {
     });
 
     expect(response.body).toContain(
-      `${defaultApp().entity.name} - Application Event`,
+      `Application ${defaultApp().entity.name} Event details`,
     );
 
     expect(response.body).toContain(
@@ -94,7 +94,7 @@ describe('application event', () => {
     });
 
     expect(response.body).toContain(
-      `${defaultApp().entity.name} - Application Event`,
+      `Application ${defaultApp().entity.name} Event`,
     );
 
     expect(response.body).toContain(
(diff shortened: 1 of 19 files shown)


tlwr (approved):

paroxp (approved):
To be truthful, the order of words makes no sense to me, but this may be my lack of language skill.

I don't want to throw it under the bus though, so if you're happy with it and perhaps a content writer also sees it OK, go ahead :)

kr8n3r:
thanks both
we'll run this past content
"Space <nameofspace>" caption above the page title "Applications" as it were does denote the "group" this page belongs to

kr8n3r:
all i got from content was "this seems like it would be a helpful change"
will merge

PR #1017, opened 2020-08-19
Update pricing calculator page with form labels and matching title

kr8n3r:
What
----

The select fields related to the ‘Estimate your monthly costs’ table not
labelled.

This means that some users of assistive technology may not be able to
determine the form elements function or purpose.

Fixes: https://www.pivotaltracker.com/n/projects/1275640/stories/174299209

This PR 
- updates title property to match page heading
- adds visually hidden labels to select fields
- removes css that was making label hidden and inaccessible to
screenreaders

**Before**
before

**After**
after

How to review
-------------

- check out branch, run locally
- got to /calculator page
- check each select has a appropriately descriptive visually hidden label and that label's `for` attribute value matches the `select`'s `id`

Who can review
---------------

not @kr8n3r

### DIFF
--- a/src/components/calculator/calculator.scss
+++ b/src/components/calculator/calculator.scss
@@ -117,16 +117,9 @@
   }
 
   select {
-    @include govuk-font($size: 16);;
+    @include govuk-font($size: 16);
   }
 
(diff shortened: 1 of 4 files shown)


paroxp (approved):
<3

PR #1036, opened 2020-08-24
[WIP] Introduce API tokens

paroxp:
What
----

We’d like to have the ability of modification permissions for users that
are not necessarily humans. PaaS Accounts requirement is restricting
this from happening.

We’re still logging, whilst not throwing any errors.

Users started to come into the support, asking for non-email based users
which could be used for their CI systems to interact with PaaS.

This feature, is simply based on the username and password system, but
relies on specific user naming, in order to differentiate between the
Users and Tokens.

An Admin and Org Manager should be able to create unlimited number of
these tokens, retreive the Key and Secret, setup the permissions for
Robot User and use the feature as they would with a google group user.

Now that we are able to have API Tokens, we don't necessarily want to
display them on the same list as the users. This could be confusing.
Filtering out by name should allow us to have the correct numbers in
place.

[WIP]

It would be great if the feature allowed for easy rotation of the
secret.

At the moment, this could be achieved by deleting the user and
re-creating. This however could be problematic when one would want to
replicate the permissions of the deleted user - especially when revoking
the token and not noting what access it used to have.

This feature however, would require either an Admin token, or the Pazmin
tool to have `uaa.admin` permissions, which it currently doesn't have.

How to review
-------------

- Sanity check the idea
- Deploy to your environment
- Attempt to use the feature
  - Create token
  - Rotate token
  - Revoke token
  - List users - see no tokens
  - List tokens - see no users

### DIFF
--- a/src/components/api-tokens/controllers.tsx
+++ b/src/components/api-tokens/controllers.tsx
@@ -255,3 +255,98 @@ export async function create(ctx: IContext, params: IParameters, body: ICreateUs
     ),
   };
 }
+
+export async function confirmRotation(ctx: IContext, params: IParameters): Promise<IResponse> {
+  const cf = new CloudFoundryClient({
+    accessToken: ctx.token.accessToken,
+    apiEndpoint: ctx.app.cloudFoundryAPI,
+    logger: ctx.app.logger,
+  });
+
+  const uaa = new UAAClient({
+    apiEndpoint: ctx.app.uaaAPI,
+    clientCredentials: {
+      clientID: ctx.app.oauthClientID,
+      clientSecret: ctx.app.oauthClientSecret,
+    },
+  });
+
+  const organization = await cf.organization(params.organizationGUID);
+
+  await checkIfAuthorised(ctx, cf, organization.metadata.guid);
+
+  const token = await uaa.getUser(params.tokenGUID);
+  if (!token || !isTokenUser(organization.entity.name, token.userName)) {
+    throw new NotFoundError('token not found');
+  }
+
+  const template = new Template(ctx.viewContext, 'Are you sure you\'d like to rotate the following API Token?');
+  template.breadcrumbs = fromOrg(ctx, organization, [
+    {
+      href: ctx.linkTo('admin.organizations.tokens.list', { organizationGUID: organization.metadata.guid }),
+      text: 'API Tokens',
+    },
+    { text: tokenName(organization.entity.name, token.userName) },
+  ]);
+
+  return {
+    body: template.render(
+      <ConfirmAction
+        action="rotate"
+        csrf={ctx.viewContext.csrf}
+        linkTo={ctx.linkTo}
+        organization={organization}
+        token={token}
+      />,
+    ),
+  };
+}
+
+export async function rotate(ctx: IContext, params: IParameters): Promise<IResponse> {
+  const cf = new CloudFoundryClient({
+    accessToken: ctx.token.accessToken,
+    apiEndpoint: ctx.app.cloudFoundryAPI,
+    logger: ctx.app.logger,
+  });
+
+  const uaa = new UAAClient({
+    apiEndpoint: ctx.app.uaaAPI,
+    clientCredentials: {
+      clientID: ctx.app.oauthClientID,
+      clientSecret: ctx.app.oauthClientSecret,
+    },
+  });
+
+  const organization = await cf.organization(params.organizationGUID);
+
+  await checkIfAuthorised(ctx, cf, organization.metadata.guid);
+
+  const token = await uaa.getUser(params.tokenGUID);
+  if (!token || !isTokenUser(organization.entity.name, token.userName)) {
+    throw new NotFoundError('token not found');
+  }
+
+  const tokenSecret = generateSecret();
+  await uaa.forceSetPassword(token.id, tokenSecret);
+
+  const template = new Template(ctx.viewContext, 'Successfully generted token secret');
+  template.breadcrumbs = fromOrg(ctx, organization, [
+    {
+      href: ctx.linkTo('admin.organizations.tokens.list', { organizationGUID: organization.metadata.guid }),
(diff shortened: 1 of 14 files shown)


src/components/api-tokens/controllers.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+  await checkIfAuthorised(ctx, cf, organization.metadata.guid);
+
+  const token = await uaa.getUser(params.tokenGUID);
+  if (!token || !isTokenUser(organization.entity.name, token.userName)) {
+    throw new NotFoundError('token not found');
+  }
+
+  await cf.setOrganizationRole(organization.metadata.guid, token.id, 'users', false);
+  await uaa.deleteUser(token.id);
```
tlwr, line 139: My preference is to remove them from the organisation but not delete the user, if we can avoid giving Pazmin `uaa.admin`.

We might just need `scim.write` if we locked the user instead. Although I think removing from the organisation should be sufficient

src/components/api-tokens/token.ts
```diff
@@ -0,0 +1,14 @@
...
+export const DEFAULT_SECRET_LENGTH = 64;
+
+export function generateKey(prefix: string, name: string, length = DEFAULT_KEY_LENGTH): string {
+  const key = randomBytes(length * 2).toString('hex').slice(0, length);
+
+  return `${prefix}-${name}-${key}`;
+}
+
+export function generateSecret(length = DEFAULT_SECRET_LENGTH): string {
+  return randomBytes(length * 2).toString('base64').slice(0, length);
```
tlwr, line 13:
```suggestion
  return randomBytes(length * 2).toString('hex').slice(0, length);
```

So we can avoid `=` `/` and `+`

src/components/api-tokens/views.tsx
```diff
@@ -0,0 +1,309 @@
...
+            draggable="false"
+            className="govuk-button"
+            data-module="govuk-button"
+          >
+            Create a new API Token
+          </a>
+        </div>
+      </div>
+
+      <h2 className="govuk-heading-m">Current API Tokens</h2>
```
kr8n3r, line 71: can we only have the table if there are any tokens?
If there are no token either have text 'There are no tokens' or better word, or not render the heading either, whichever one is easietr

src/components/api-tokens/views.tsx
```diff
@@ -0,0 +1,309 @@
...
+          <span id="name-hint" className="govuk-hint">
+            Must be a combination of lowercase alphanumeric characters separated by hyphen (<code>-</code>).
+          </span>
+          {props.error
+            ? <span id="name-error" className="govuk-error-message">
+                <span className="govuk-visually-hidden">Error:</span>{' '}
+                API Token name needs to be a combination of lowercase alphanumeric characters separated by hyphen.
+              </span>
+            : <></>}
+          <input
```
kr8n3r, line 272: don't think we should rely on browser input validation?

src/components/api-tokens/controllers.tsx
paroxp: The need for `uaa.admin` is not to delete the user, but to "force password change". I know we discussed it, but I think without that functionality the feature may be problematic to users.

As I've described in the PR comment:

> This however could be problematic when one would want to
replicate the permissions of the deleted user - especially when revoking
the token and not noting what access it used to have.

Essentially, I'm worried this would be more appealing feature and I'm not certain if there is much sense delivering these `API Tokens` when we rotate via `Delete user` followed by `Create user` followed by `Set permissions`.

---

But to address your comment, yes, we can definitely replace `deleteUser` with `detachUser` and get similar behaviour.

src/components/api-tokens/token.ts
paroxp: Damn, I started with hex, but wanted some characters in the token xD

kr8n3r:
sorry only saw now, this is an RFC

src/components/api-tokens/controllers.tsx
LeePorte: Password changing is going to be a pretty essential feature. I'd also like to see something in place for handling dead accounts, lest there be a large amount of cruft users that exist in UAA never to be resurrected.

TBH I'm less concerned about giving Pazmin uaa.admin as we are going to be playing an easy on easy off trial accounts epic in the next quarter where if you have the appropriate email domain you can set up a trial account without the involvement of support, 90 days later that org and user are removed automagically.

46bit:
I'm thinking about this in light of my https://github.com/alphagov/paas-prometheus-endpoints work. It takes a CF username/password and fetches metrics for all the service instances that user can see.

The current instructions would be along the line of "make a user and give it space auditor on every space you want metrics for" (irritatingly, org auditor doesn't work.)

If we have API tokens/service users like this it'd remove the need to make a user, and all that email validation. We could also have a little page for setting up specific metrics users. They'd still be exposed as Space Auditors in the interface but people wouldn't have to understand quite so much.

Until/unless we get more users for https://github.com/alphagov/paas-prometheus-endpoints this is wildly theoretical.

paroxp:
In response to @46bit 

Something about the process of users being able to create other users rings alarm bells in my head.

I think it may be easier to loose control of such user. For instance no ability to reset-passwords (due to lack of email). This could cause a security risks...

This PR, fixes that with a `Rotate` and `Revoke` options, as well as enforces usernames and passwords to some standard, which not only allow us to divide tokens from users, but also would allow easier audit logs monitoring. Imagine:

> myMadeUpUsername viewed secrets for the www app in production space

Who's to say if it's a robot, or a user? Thinking from the perspective of our own Cyber Security team.

I am indifferent to be honest :) Just wouldn't want to confuse things further - and `fake drawn boxes on piece of paper` prove to be efficient barrier for understanding :trollface:

46bit:
I don't think we're on the same page, @paroxp. I was mostly saying how wonderful this PR will be to have merged.

PR #1039, opened 2020-08-25
Update org manage quota page

kr8n3r:
What
----

- update heading and title to be consistent with https://github.com/alphagov/paas-admin/pull/1016
- update quota `select` label
- associate select and hint with `aria-describedby`

**Before**

before

**After**

after

How to review
-------------

- deploy to env (doesn't work in local dev)
- go to org -> manage quota
- check title and page heading match (page heading has visually-hidden text "organisation)
- check select `aria-describedby` attribute value matches hint's `id`

Who can review
---------------

not @kr8n3r


paroxp (approved):
Nice. Sorry about that. I was under the impression that if this page is entirely for platform operators and not our user base, it may not require such detail attention. It doesn't hurt nevertheless.

Thanks!

PR #1041, opened 2020-08-25
Billing statements page updates

kr8n3r:
What
----
_Note: will squash commits after review_

Problem:

The table contains form elements that are not descriptive enough for screen reader users to determine that their purpose is to filter the options in the corresponding column:
• Name
• Space
• Plan
• Inc VAT

Ensure that the table column/buttons are uniquely descriptive of their function or purpose and enable users to identify which column/button is currently selected (i.e. Filter by ‘Name’ selected).
Also, users need to be informed that the filter has been applied.

Solution:
We have address above with the following:

- Update statements title & heading to reflect selected filters
- Add visually hidden "Sort by" text to table sort column buttons
- Provide a visually-hidden table caption
- Add `aria-sort` attribute to table columns
- add a visually-hidden caption to cost summary table

How to review
-------------
Suggest reviewing by commit
- go to org billing page
- apply filters and sort table by a column
- inspect page title and h1 for text that reflects selected filters and sort option
- inspect table caption for text that reflects selected filters and sort option
- inspect table header (`th`). only the selected filter should have `aria-sort` attribute

### Heading and title reflecting selected filters and sort order
Screenshot 2020-08-26 at 09 08 45

### Table caption read by screen reader before and after

Screenshot 2020-08-25 at 11 26 49
Screenshot 2020-08-25 at 11 27 10

### Updated sort button text showing descriptive text and sort direction
button VO text

### Cost summary table caption
Screenshot 2020-08-26 at 10 11 51
 

Who can review
---------------

not @kr8n3r

### DIFF
--- a/src/components/statements/_statements.scss
+++ b/src/components/statements/_statements.scss
@@ -78,10 +78,6 @@
   @include hidden;
 }
 
-.paas-billing-heading {
-  margin-bottom: 20px;
-}
-
 .paas-table-notification {
   @include govuk-font($size: 24);
   padding: 25px 0;
--- a/src/components/statements/_statements.scss
+++ b/src/components/statements/_statements.scss
@@ -53,14 +53,20 @@
     &::after {
       display: inline-block;
       margin-left: .5rem;
+      content: "";
     }
 
-    &.asc::after {
-      content: "\25b2";
+    // use border for trianglr shape instead of content
+    // as content get's read by screenreaders
+
+    &.asc:after {
+      @include govuk-shape-arrow($direction: up, $base: 12px, $display: inline-block);
+      border-bottom-color: currentColor;
     }
 
-    &.desc::after {
-      content: "\25bc";
+    &.desc:after {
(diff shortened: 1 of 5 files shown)


46bit:
Out of interest, what does `{' '}` do?

46bit:
I've gone through the commits and this looks good. Want to check it in a browser next week.

kr8n3r:
> Out of interest, what does `{' '}` do?
It adds a space between nodes
in pazmin we have a test `spacesMissingAroundInlineElements` that will fail on `<span> <span>` but pass on `<span>{' '}</span>` for example

paroxp:
It's react's rendering engines flaw IMO, which basically attempts to minimise the syntax by removing spaces it claims unnecessary. So `Look at <span>this</span>.` Would read `Look atthis.`. As Jani mentioned, adding `{' '}` will tell react that there is a programatic and not just useless space, which could have been added by ones editor.

paroxp (approved):

PR #1044, opened 2020-08-27
Keyboard accessible <abbr>eviation expansion

kr8n3r:
**Note: will squash commits after review**

What
----

**Problem**: 
The acronyms such as (MiB/KIB/GiB and others) are expanded with mouse hover, however, this functionality is not accessible to standard keyboard commands. This means that some users who are unfamiliar with the acronym and are not able to access its expanded form may not understand its purpose or meaning.

**Solution**
Create a keyboard tabbable and dismissible tooltip that expands the acronym. 

Note about WCAG 2.1 Success Criterion 1.4.13 Content on Hover or Focus:
This success criterion advises that if anything is covering any other content on mouse hover
or keyboard focus, it must be ‘dismissible’ without moving the mouse or keyboard focus.
(e.g. by using the ‘Esc’ key)

To achieve this, we modified the `abbr` to no longer have the description in the tittle attribute. Instead, we add an `aria-label` attribute that holds that description which becomes visible as a "tooltip" on `:after` pseudo on keyboard focus or mouse hover of `abbr`.
To make it keyboard focusable, we added `tabindex=0` to `abbr` element.

Additionally it requires so javascript that listens to events on the target element and makes the expanded description visible. If Javascript is disabled, we do not show the tooltip, but the description is still accessible via keyboard.

Fixes: https://www.pivotaltracker.com/n/projects/1275640/stories/174195569

How to review
-------------

- best to review commit by commit
- go to any page with acronyms
- use mouse to hover of the element. 
  - Tooltip should appear/disappear
- use keyboard TAB key to tab to the element
  - on focus tooltip should appear
  - it should be dismissed with ESC key
  - it should also disappear when focus moves away

Who can review
---------------

not @kr8n3r 

Visual changes
----

### On mouse hover
Screenshot 2020-08-28 at 08 16 41

### On keyboard focus
Screenshot 2020-08-28 at 08 16 50

### Interactive example showing hover/focus/dismiss
tooltip interactive example

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -63,6 +63,7 @@
     "import/no-commonjs": "error",
     "import/order": ["warn", {"alphabetize": {"order": "asc", "caseInsensitive": true}, "newlines-between": "always"}],
     "jsx-quotes": ["error", "prefer-double"],
+    "jsx-a11y/no-noninteractive-tabindex": "warn",
     "jsx-a11y/no-redundant-roles": ["error", {"details": ["group"]}],
     "linebreak-style": ["error", "unix"],
     "max-len": ["warn", {"code": 120}],
--- a/src/components/marketplace/views.tsx
+++ b/src/components/marketplace/views.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 
 import { CommandLineAlternative, NoTick, Tick } from '../../layouts/partials';
+import { Abbreviation } from '../../layouts/helpers';
 import { IV3Service, IV3ServicePlan } from '../../lib/cf/types';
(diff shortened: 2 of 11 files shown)


paroxp (commented):
I wonder, if this doesn't call for react component instead?

```jsx
High {abbrTooltip('IOPS','Input/Output Operations Per Second')}
```
vs
```jsx
High <Abbreviation acronym="IOPS">Input/Output Operations Per Second</Abbreviation>
```

kr8n3r:
updated to use a React component

paroxp (approved):

PR #1055, opened 2020-09-03
Add number of service instances to spaces list

46bit:
What
----

This adds the number of backing services to the list of spaces:

Screenshot 2020-09-03 at 17 47 09

Without this change, the info isn't available to our users without them doing manual clicking or learning to interact with the CC API. With this change it is very easy to see where the service instances are.

As you can see from the screenshot this looks good even on narrow displays.

How to review
-------------

Happy with the change?
Code review.

Who can review
---------------

Not @46bit.

### DIFF
--- a/src/components/applications/views.tsx
+++ b/src/components/applications/views.tsx
@@ -3,7 +3,7 @@ import React, { ReactElement, ReactNode } from 'react';
 import { CommandLineAlternative } from '../../layouts/partials';
 import { IApplication } from '../../lib/cf/types';
 import { RouteActiveChecker, RouteLinker } from '../app';
-import { IEnchancedApplication } from '../spaces/views';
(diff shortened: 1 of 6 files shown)


tlwr:
I reckon if we are worried about visual clutter we could remove the "stopped apps" column

review in progress

46bit:
Sorry, the delay before Concourse starts running the tests made me think the testsuite already passed. I need to add a mock.

tlwr:
LGTM when the aforementioned mock is added :)

46bit:
I think the *Stopped apps* column is an alright idea but what's actually needed is a sense of any apps that are unstable (e.g., crashing often.)

tlwr:
Separately, we could add the number of services to the quotas blurb
image

PR #1059, opened 2020-09-10
add sqs service marketplace metadata

chrisfarms:
What
----

adds the service icon for the upcoming SQS service alpha so it looks nice when it appears in pazmin marketplace listings for orgs where it is enabled.

How to review
-------------

Is that the right icon?

Who can review
---------------

Not @chrisfarms


tlwr (approved):

kr8n3r:
any links to docs, etc?

chrisfarms:
> any links to docs, etc?

GOV.UK PaaS docs are pending ... this service is initially not exposed to all tenants while we preview it privately

Links to AWS SQS service docs are passed through from the service plan data exposed by the marketplace/api, so should automagically appear once the service appears.

kr8n3r:
👍

PR #1060, opened 2020-09-10
Mark suspended orgs

AP-Hunt:
What
----
Org in CloudFoundry have two states: `active` and `suspended`. The former is the default state.

We expect to start moving orgs to the `suspended` state when we want to shut them down. Currently it is only possible to know an org is suspended by using the API directly. This PR marks suspended orgs in the UI.

image

image

How to review
-------------
Deploy it, and suspend an org

```
cf curl -X PATCH -d '{"suspended": true}' -H "Content-Type: application/json" "/v3/organizations/$(cf org ORG_NAME --guid)"
```

Who can review
---------------
Anyone


46bit (approved):
LGTM

kr8n3r (approved):

PR #1068, opened 2020-09-17
add backing service metrics  for number of sqs messages sent/recv

chrisfarms:
## What

display charts for number of messages sent/recv for sqs backing services

this is not all the metrics available, it was an initial pass to get the ball rolling and ensure everything is in place to add metrics as specific needs arise.

currently only checks the "primary" queue (ignore the "secondary" queue that is often configured as a dead-letter queue).

## How to review

If you have an `aws-sqs-queue` service instance  you should be able visit the metrics tab and see charts for NumberOfMessagesSent and NumberOfMessagesReceived

## Who can review

Not @chrisfarms

### DIFF
--- a/src/components/service-metrics/controllers.tsx
+++ b/src/components/service-metrics/controllers.tsx
@@ -16,6 +16,8 @@ import {
   elasticsearchMetricNames,
   RDSMetricDataGetter,
   rdsMetricNames,
+  SQSMetricDataGetter,
+  sqsMetricNames,
 } from '../../lib/metric-data-getters';
 import roundDown from '../../lib/moment/round';
 import PromClient from '../../lib/prom';
@@ -30,6 +32,7 @@ import {
   elastiCacheMetrics,
   elasticSearchMetrics,
   rdsMetrics,
+  sqsMetrics,
 } from './metrics';
 import { getPeriod } from './utils';
 import {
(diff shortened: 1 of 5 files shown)


chrisfarms:
Note: since the revelations that FIFO queues have weird naming restrictions, we will need to change the logic that generates the correct QueueName dimension for the metrics to produce special names for "fifo" plans after the broker is updated :disappointed: .... we can fix this in a follow up PR though, this one will work fine for standard queues

paroxp:
did you happen to deploy it to your env and have a service you could link to?

chrisfarms:
> did you happen to deploy it to your env and have a service you could link to?

No sorry, I've been running locally against my dev env

chrisfarms:
send-msg-chart_002

Here's a really unexciting graph with a little blip of queue usage if that helps

src/components/service-metrics/controllers.tsx
```diff
@@ -398,6 +422,33 @@ export async function downloadServiceMetrics(
...
+        rangeStart,
+        rangeStop,
+        servicePlan.entity.name,
+      );
+
+      headers = ['Service', 'Time', 'Value'];
+      contents = values(sqsMetricSeries[params.metric])
+        .map(metric =>
+          metric.metrics.map(series => [
+            serviceLabel,
```
paroxp, line 445: I think this is wrong, but it's not your fault (you copy pasted)...

It will essentially create a column full of `aws-sqs-queue` which seems pointless.

I am not certain if it should be a `service.name` which also would be a massive duplication, or if this column should/could be deleted xD

I'll bring it up with our PMs.

paroxp (approved):

src/components/service-metrics/controllers.tsx
chrisfarms: yeah I thought it was bit weird ... kinda makes sense if you download all the CSVs and merge the data together or something :shrug:

PR #1071, opened 2020-09-18
Fix accounts usage for autoscaler

46bit:
What
----

The app and space audit event pages are erroring when they include audit events from the [autoscaler](https://github.com/cloudfoundry/app-autoscaler). This is happening because we never anticipated this sort of actor field on an audit event:

```
    "actor": {
      "type": "user",
      "guid": "app_autoscaler",
      "name": ""
    },
```

We assumed that anything with `type: user` would have a UUID for a GUID. This makes complete sense, but it isn't the case for the autoscaler.

The autoscaler uses a UAA Client named `app_autoscaler`. It is reported with `type: user` even though it's a UAA Client, not a UAA user. The UUID is set to the name of the client, not a GUID.

Our views already had logic to cope with this problem. They fall back to printing out `<code>{{event.actor.guid}}</code>`. But the controllers try to convert GUIDs to validated user emails using https://github.com/alphagov/paas-accounts, and that code did not guard against this problem.

This PR changes the controllers so they only ask https://github.com/alphagov/paas-accounts about actors whose UUID is a valid GUID.

How to review
-------------

* <s>Deploy to a dev env</s> I used a modified stub API to test this
* [x] Check this works fine with normal events
* [x] Look at events on the autoscaler's test app
* [x] Look at events on the space containing the autoscaler's test app

Who can review
---------------

Not @46bit

### DIFF
--- a/src/components/application-events/controllers.test.tsx
+++ b/src/components/application-events/controllers.test.tsx
@@ -2,12 +2,11 @@ import lodash from 'lodash';
 import moment from 'moment';
 import nock from 'nock';
 
-
 import { DATE_TIME } from '../../layouts';
 import { spacesMissingAroundInlineElements } from '../../layouts/react-spacing.test';
 import * as data from '../../lib/cf/cf.test.data';
 import { app as defaultApp } from '../../lib/cf/test-data/app';
-import { auditEvent as defaultAuditEvent } from '../../lib/cf/test-data/audit-event';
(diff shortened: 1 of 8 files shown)


46bit:
I've tested this locally rather than in a dev env, and checked every relevant page. Here are two examples:

Screenshot 2020-09-18 at 15 35 23

Screenshot 2020-09-18 at 15 35 27

AP-Hunt (approved):
Looks smart enough to me 👍

PR #1072, opened 2020-09-18
Run tests on GitHub Actions instead of our own CI

AP-Hunt:
What
----

The tests we run don't require any special resources or permissions, and in the
past they've got stopped the PR being merged when they should have. Moving to
GitHub Actions ought to make the whole thing a bit more reliable.

How to review
-------------

See the Actions tab here, and ensure the tests passed.

Who can review
---------------
Anyone but me

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -1,0 +1,25 @@
+---
+name: pr-tests
+on: pull_request
+
+jobs:
+  test:
+    name: Run tests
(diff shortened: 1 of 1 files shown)


AP-Hunt:
> We'll need to disable the tests being run by `paas-release-ci` too :)

Agree. That's the next job. I can't remove the `ci/` directory until I've removed the pipeline in `paas-release-ci`. I'll do that in a second PR.

PR #1073, opened 2020-09-20
Use tabs to show trial and billable organisations on the admin organisation report

kr8n3r:
What
----

Move the display of trial and billable organisations into tabs layout - true, arrow key navigable tabs as per [Design system](https://design-system.service.gov.uk/components/tabs/)

On Production the list of both is long this makes it easier to only view either.

Additionally, we now only show a table of organisations actually exist.

How to review
-------------

- deploy or run locally
- visit /reports/organisations
- check that tab work as expected (as per DS)

Who can review
---------------

not @kr8n3r 

Visual change
-----------

### Before
Screenshot 2020-09-20 at 08 13 39

### After
Screenshot 2020-09-20 at 08 10 16
Screenshot 2020-09-20 at 08 10 23

### DIFF
--- a/src/frontend/javascript/init.js
+++ b/src/frontend/javascript/init.js
@@ -44,3 +44,11 @@ if ($tooltips) {
     new Tooltip($tooltips[i]).init();
   };
 }
+
+var $tabs = document.querySelectorAll('[data-module="govuk-tabs"]');
+var GOVUKTabs = window.GOVUKFrontend.Tabs;
+if ($tabs) {
+  for (var i = 0; i < $tabs.length; i++) {
+    new GOVUKTabs($tabs[i]).init();
+  };
+}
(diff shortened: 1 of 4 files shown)


src/frontend/javascript/init.js
```diff
@@ -44,3 +44,11 @@ if ($tooltips) {
     new Tooltip($tooltips[i]).init();
   };
 }
+
+var $tabs = document.querySelectorAll('[data-module="govuk-tabs"]');
```
paroxp, line 48: Do you know if this will break our other tabs?

`data-module="govuk-tabs"` is used here: https://github.com/alphagov/paas-admin/blob/693ba3e7049c441a0f84ace817bfbcf88e39ef1b/src/components/services/views.tsx#L80

paroxp: Nevermind, just noticed you're fixing it.

kr8n3r: yep. it wouldn't break, just throw a js error.

PR #1091, opened 2020-10-15
Add mailing list opt-in

paroxp:
There are some linting fixes, safe to skim read them.

## How to review

- Deploy pazmin to your dev env
- `cf set-env MAILCHIMP_API_KEY`
- Run through `/support/sign-up` process - opt into mailing list

Screenshot 2020-10-15 at 10 06 27


kr8n3r (approved):
it looks good, can't test it works
subject to changing the words of
>Yes, sign me up to your glorious mailing list

src/components/support/controllers.test.tsx
```diff
@@ -122,6 +122,34 @@ describe(controller.HandleSignupFormPost, () => {
     expect(response.body).toContain('We have received your request');
   });
 
+  it('should create a zendesk ticket correctly whilst signing user up for mailing list', async () => {
```
46bit, line 125: This test name doesn't make sense. Do we really want to create a zendesk ticket? It looks like we sign them up to Mailchimp instead.

paroxp: The signup to mailchimp, is part of the sign up process. It's an extra box that can be ticked when requesting a paas account.

46bit: OK. We probably need this feature available within the pazmin interface too. 🤔

46bit:
Mailchimp's API returned a `404` when I tried to use this in my dev environment. Waiting to get Mailchimp access so I can try and debug.

46bit:
With the latest commit and https://github.com/alphagov/paas-credentials/pull/183 this seems to talk fine to Mailchimp. But I am not receiving any invite emails, and I can't see the new users anywhere in the Mailchimp UI. Will need to dig into that next week.

paroxp:
Oof, I was using the web-id... Nice catch.

This is annoying. I just ran few tests with CURL, to add myself. Setting the user to `pending`, does not display them in the UI, nor does it email the confirmation to the user - which the documentation claims it would at places...

It does however list them in the API view - which was the initial thing I was testing. This then enables us to modify their status to `subscribed`. Almost as it's now expected to run our own validation?

I have then attempted to use some stack overflow magic, to find out that there is an option for entire "list", to allow double-opt-in... Which also doesn't work. But not only that, there is an option I've ticked, to send out final welcome email to any new subscribers - AND THAT DOESN'T WORK EITHER...

Final test, was using the UI. No changes there.

I am a little bit confused if I'm honest, as to why the only thing we're asking for (sending emails) is not being performed...

paroxp:
Just in my defence and for documentation purposes:

https://mailchimp.com/developer/guides/create-your-first-audience/#add-a-contact-to-an-audience

> Use `pending` to send a confirmation email. Once confirmed, the contact’s status will update to `subscribed`.

46bit:
It could be using "Use pending to send a confirmation email" in the passive tense. Describing a usage flow rather than saying it will do things for us. But I don't know about that--I'm sure I've received signup confirmation emails from Mailchimp before.

paroxp:
> I'm sure I've received signup confirmation emails from Mailchimp before.

Same here.

PR #1105, opened 2020-11-06
Update marketplace service docs text logic

kr8n3r:
What
----

Our user pointed out that on
https://admin.london.cloud.service.gov.uk/marketplace/754c7e96-bc62-45ca-8fe7-a5064d922588
it should say ElastiCache instead of RDS.

Screenshot 2020-11-06 at 14 59 21

Our url text logic was only checking for hostname which is docs.aws.amazon.com for both AWS ElastiCache Redis and AWS RDS Postgres.

Updated logic checks for both hostname and pathname to return the correct link text.

How to review
-------------

deploy to a dev env as this doesn't work locally

Who can review
---------------

not @kr8n3r

### DIFF
--- a/src/components/marketplace/views.tsx
+++ b/src/components/marketplace/views.tsx
@@ -177,6 +177,10 @@ function documentationTitle(service: string, url: string): string {
       return `GOV.UK PaaS ${service} documentation`;
     case 'aws.amazon.com':
     case 'docs.aws.amazon.com':
+        switch(true){
+          case u.pathname.toLowerCase().includes('elasticache'):
+            return `AWS ElastiCache ${service} documentation`;
+        }
       return `AWS RDS ${service} documentation`;
     case 'help.aiven.io':
       return `Aiven ${service} documentation`;


src/components/marketplace/views.tsx
```diff
@@ -171,14 +171,17 @@ const serviceDetails: IServiceDetails = {
 /* istanbul ignore next */
 function documentationTitle(service: string, url: string): string {
   const u = new URL(url);
-
-  switch (u.hostname) {
-    case 'docs.cloud.service.gov.uk':
+  const hostName =  u.hostname;
```
AP-Hunt, line 174: I like this approach, but I think it'd be more readable as

```
swtich(hostName){
    case "docs.cloud.service.gov.uk": return "PaaS"
    case "aws.amazon.com": 
    case "docs.aws.amazon.com":
        switch(true){
            case pathName.toLowerCase().includes("elasticache"): return "Elasticache"
        }
}
```

kr8n3r: updated but unable to test

PR #1106, opened 2020-11-09
Correct link text for AWS service documentation

AP-Hunt:
What
----
We were giving the title "AWS RDS ${service} documentation" to every link to AWS docs. 
This resulted in the Redis docs being labelled "AWS RDS Redis documentation".

This commit makes the link text a bit more generic, so that it can apply to all links to AWS docs.

How to review
-------------
See if you agree with the text

Who can review
---------------
Anyone

### DIFF
--- a/src/components/marketplace/views.tsx
+++ b/src/components/marketplace/views.tsx
@@ -177,7 +177,7 @@ function documentationTitle(service: string, url: string): string {
       return `GOV.UK PaaS ${service} documentation`;
     case 'aws.amazon.com':
     case 'docs.aws.amazon.com':
-      return `AWS RDS ${service} documentation`;
+      return `AWS service documentation`;
     case 'help.aiven.io':
       return `Aiven ${service} documentation`;
     default:


src/components/marketplace/views.tsx
```diff
@@ -228,7 +228,6 @@ function TableRow(props: ITableRowProperties): ReactElement {
       {props.limitsConcurrentConnections
         ? <td className="govuk-table__cell govuk-table__cell--numeric">
             {props.data.concurrentConnections}
-          </td>
```
paroxp, line 231: probably unintended - will cause errors

paroxp (changes requested):

src/components/marketplace/views.tsx
AP-Hunt: Good catch

kr8n3r:
preferred over #1105?

AP-Hunt:
> preferred over #1105?

Oh, sorry! I didn't know that existed. I will review that and see if it solves this issue.

kr8n3r:
no, this one is more generic

AP-Hunt:
Preferring #1105

PR #1115, opened 2020-11-23
Fix missing mou pdf files

kr8n3r:
What
----

Files are located in /dist/downloads folder and were deleted before every build by the `npm run clean` command.

That command used `rm -rf dist/*` which wiped everything.

It has now been updated with `find` to exclude the 'downloads' folder.

PDF file names have been prefixed with 'govuk-paas' and download links
amended.

How to review
-------------

- checkout branch
- run locally with `npm run start:stub-api` and `npm run start:with-stub-api` to run the app.
- link links to MOUs in the footer. 
- each pdf should download
- check `dist/downloads`  - pdf files should be there, no git changes
- review code

Who can review
---------------

not @kr8n3r


schmie (approved):
:+1:

PR #1124, opened 2020-12-14
marketplace: Update logos

paroxp:


kr8n3r (approved):
I've resized the sqs logo :)

PR #1132, opened 2021-01-25
Replace ES6 with ES5 in cookie functions

paroxp:
Original: #1126

What
----

No arrow functions allowed in client-side JS

How to review
-------------

- set `this.cookieDomain` to empty string for localhost test in cookie-functions.js
- start the app
- save a cookie preference
- check if have cookies stored (chrome: Option + ⌘ + J (on macOS), or Shift + CTRL + J (on Windows/Linux) - Application tab

Who can review
---------------
not @kr8n3r


kr8n3r (approved):
tested again, no change in functionality

PR #1135, opened 2021-01-25
Move from node-sass to sass

paroxp:
Original: #1095

What
----

Node-sass is being [deprecated](https://sass-lang.com/blog/libsass-is-deprecated) (node-sass is a wrapper for Libsass) so we need to move to the recommended sass (dart package).

For precedent, were already using sass compiler with govuk-frontend in the [product pages build](https://github.com/alphagov/paas-product-pages/blob/master/package.json#L26)

How to review
-------------

- checkout branch, build it
- look around the pages
- there should be no visual changes or broken things

Who can review
---------------

not @kr8n3r


kr8n3r (approved):
tested locally, no visible changes as expected

PR #1136, opened 2021-01-25
Update marketplace service docs text logic

paroxp:
Original: #1105

What
----

Our user pointed out that on
https://admin.london.cloud.service.gov.uk/marketplace/754c7e96-bc62-45ca-8fe7-a5064d922588
it should say ElastiCache instead of RDS.

Screenshot 2020-11-06 at 14 59 21

Our url text logic was only checking for hostname which is docs.aws.amazon.com for both AWS ElastiCache Redis and AWS RDS Postgres.

Updated logic checks for both hostname and pathname to return the correct link text.

How to review
-------------

deploy to a dev env as this doesn't work locally

Who can review
---------------

not @kr8n3r

### DIFF
--- a/src/components/marketplace/views.tsx
+++ b/src/components/marketplace/views.tsx
@@ -177,6 +177,10 @@ function documentationTitle(service: string, url: string): string {
       return `GOV.UK PaaS ${service} documentation`;
     case 'aws.amazon.com':
     case 'docs.aws.amazon.com':
+        switch(true){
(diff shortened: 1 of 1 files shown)


kr8n3r:
tested on staging

change visible
Screenshot 2021-03-05 at 14 50 18

no affect to others
Screenshot 2021-03-05 at 14 49 47

kr8n3r (approved):

PR #1137, opened 2021-01-25
Move from momentjs to datefs

paroxp:
Original: #1096

What
----

Moment.js is now in [maintenance mode only](https://momentjs.com/docs/#/-project-status/) so we can take time to move away from it.

[date-fns](https://github.com/date-fns/date-fns) is a much smaller replacement. We utilise some things that are not available natively in javascript so cannot use that.
[See the comparison table](https://github.com/you-dont-need/You-Dont-Need-Momentjs#feature-parity)

How to review
-------------

Describe the steps required to test the changes.

Who can review
---------------

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7396,6 +7396,11 @@
         "whatwg-url": "^7.0.0"
       }
     },
+    "date-fns": {
+      "version": "2.16.1",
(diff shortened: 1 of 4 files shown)


kr8n3r:
closing as stale

PR #1138, opened 2021-01-25
Org report layout update

paroxp:
Original: #1073

What
----

Move the display of trial and billable organisations into tabs layout - true, arrow key navigable tabs as per [Design system](https://design-system.service.gov.uk/components/tabs/)

On Production the list of both is long this makes it easier to only view either.

Additionally, we now only show a table of organisations actually exist.

How to review
-------------

- deploy or run locally
- visit /reports/organisations
- check that tab work as expected (as per DS)

Who can review
---------------

not @kr8n3r 

Visual change
-----------

### Before
Screenshot 2020-09-20 at 08 13 39

### After
Screenshot 2020-09-20 at 08 10 16
Screenshot 2020-09-20 at 08 10 23

### DIFF
--- a/src/components/reports/views.test.tsx
+++ b/src/components/reports/views.test.tsx
@@ -223,6 +223,46 @@ describe(OrganizationsReport, () => {
 
     expect($('table tbody tr td:first-of-type span.govuk-tag').length).toEqual(1);
   })
+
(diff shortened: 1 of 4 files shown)


kr8n3r:
closing as it doesn't meet user needs

PR #1142, opened 2021-01-28
use service/app specific command for CLI example

gidsg:
What
----

Use Service/App specific command in CLI example. This is because:
1) As this is an app/service specific page it makes more sense to show the specific command
2) The call to action says _"You can also view this information in the Cloud Foundry command line interface by running:"_ and this is simply not true, the information on the `cf apps`/`cf services` summary does not contain all the fields in this view.

How to review
-------------

Go to Application and Service pages. Commands should be app/service specific
e.g.
Screenshot 2021-01-28 at 15 59 42

Who can review
---------------

not me

### DIFF
--- a/src/components/applications/views.tsx
+++ b/src/components/applications/views.tsx
@@ -222,9 +222,7 @@ export function ApplicationPage(
             </tbody>
           </table>
           </div>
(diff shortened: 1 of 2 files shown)


kr8n3r (approved):
thanks @gidsg

PR #1144, opened 2021-02-03
Update node-jose dependency

kr8n3r:
What
----

Release updates the node-forge dependency to resolve a security issue.

We're using `node-jose` in our test file src/components/account/oidc.test.ts

How to review
-------------

tests pass

Who can review
---------------

not @kr8n3r


AP-Hunt (approved):
I see the tests pass 👍

PR #1162, opened 2021-02-10
Introduce scrapping mechanism

paroxp:
What
----

We'd like to save some burden from prometheus, as pazmin constantly
performing queries against it could having an impact against the box's
memory.

Instead, we're introducing the scraper, which will be run once a day by
our pipeline, to generate JSON data, then reachable by the application
itself.

How to review
-------------

- Run the script yourself
- Serve generated file
- View performance dashboard


kr8n3r (approved):
sounds a sensible way to not blow up prometheus and have a fast-serving page.
tested locally and made a minor tweak by removing a wrapper.

ship it, subject to passing tests


Screenshot 2021-02-10 at 17 18 37

PR #1180, opened 2021-02-23
GOV.UK PaaS - Performance dashboard - dynamic uptime

paroxp:
What
----

Add integration with pingdom to pull down the downTime information and reflect on the performance page as a percentage.

How to review
-------------

- Code review
- Run the script

### DIFF
--- a/src/components/performance/scraper.ts
+++ b/src/components/performance/scraper.ts
@@ -5,15 +5,19 @@ import { IMetricSerie } from '../../lib/metrics';
 import PromClient from '../../lib/prom';
 
 export const now = moment().subtract(1, 'hour').toDate();
+const DELAY = 2000;
 export const period = moment.duration(1, 'week');
 export const timeAgo = moment().subtract(1, 'year').toDate();
 
 const delay = async (ms: number): Promise<object> => await new Promise(resolve => setTimeout(resolve, ms));
 
-interface IPromConfig {
-  readonly endpoint: string;
-  readonly password: string;
-  readonly username: string;
+interface IConfig {
+  readonly pingdom: object;
+  readonly prometheus: {
+    readonly endpoint: string;
+    readonly password: string;
+    readonly username: string;
+  };
 }
 
 export interface IScrapedData {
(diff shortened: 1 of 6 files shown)


src/components/performance/scraper.ts
```diff
@@ -36,36 +58,84 @@ const queries = {
   serviceCount: 'sum(group by (service_instance_id) (cf_service_instance_info{last_operation_type=~"create|update"}))',
 };
 
-export async function scrape(promConfig: IPromConfig, logger: BaseLogger): Promise<IScrapedData> {
+function calculateUptime(data: IPingdomUptimeResponse): number {
+  const { totalup, totaldown, totalunknown } = data.summary.status;
+  const total = totalup + totaldown + totalunknown;
+  const totalDownTime = totaldown + totalunknown;
+  const uptimePercentage = ((total - totalDownTime) / total) * 100;
```
kr8n3r, line 65: quick maths from the pingdom api gives me 5.23%? 
https://user-images.githubusercontent.com/3758555/108873938-90191e00-75f3-11eb-91d3-00e62ba7f53c.png

paroxp: Interesting, which "checkID" are you running against?

I've only tested ireland if i'm being honest - didn't think it would be "much" different from london... [Here are the three checkIDs we should care about.](https://github.com/alphagov/paas-cf/pull/2599/files#diff-3ba3763df7e55cc5ecce09f7e76a99063313c7103904fb74bd634f5f7e011e50R3991-R3997)

paroxp: Another possibility, is that we should ignore `totalunknown` in our calculations... It doesn't definitely say it the API was down, but neither does it say it was up...

paroxp: And lastly, I noticed your API response says: `from: 0` meaning ALL data since beginning of unix time.

paroxp: If you do the following:

```
export PINGDOM_API_TOKEN: ((pingdom_api_token))
export PINGDOM_CHECK_ID="3364261"
npm run -s scrape:performance
```

you will get errors for prometheus, but at least you should get a percentage right :thinking:

kr8n3r: i ran against 4685882 which does have a high amount of "unknown" and did maths by hand so very likely made a mistake somewhere. 

will run the above scrape

kr8n3r (approved):
`{"uptime":99.99}% `

PR #1423, opened 2021-05-04
Implement GA event tracking

kr8n3r:
What
----

To measure engagement for users that have accepted cookies we want to
record certain events with either specific labels for targeted elements
or more generic for larger groups .

This script introduces functions to send events for:

- targeted elements with data attributes
- external links to non *.cloud.service subdomains and email links
- error pages (403, 404, 500)

For targeted elements values for the event are read from the data
attributes.

For external links, values are partially set and partially read from
the target element.

For error pages, values are set.

This is an [adapted version of product page work](https://github.com/alphagov/paas-product-pages/pull/120)

How to review
-------------
- best to review commit by commit
- can test this with the Omnibug extension on https://admin.dev02.dev.cloudpipeline.digital for 
    top nav, external links, error page and billing page download and filter

Who can review
---------------

not @kr8n3r 
---

### DIFF
--- a/src/frontend/javascript/event-tracking.js
+++ b/src/frontend/javascript/event-tracking.js
@@ -1,0 +1,110 @@
+function EventTracking () {
+  this.$bodyElement = document.querySelector('body')
+  this.$mainHeading = this.$bodyElement.querySelector('h1')
+ }
+
+EventTracking.prototype.init = function() {
+
+  this.$bodyElement.addEventListener('click', this.handleClickEvent.bind(this))
+  this.checkForErrorPage(this.$mainHeading)
+}
+
+EventTracking.prototype.handleClickEvent = function (e) {
+  var target = e.target,
+      isLink = e.target.nodeName === 'A',
+      // external link: starts with http(s) or mailto and does not contain a "service.gov" in url
+      externalLinkRegex = /^(?=.*(http|mailto))(?:(?!cloud\.service\.gov\.uk).)*$/g,
+      isExternalLink = isLink && externalLinkRegex.test(target.getAttribute('href')),
+      // we have data tracking attributes on some elements
+      isManuallyTargetedElement = target.hasAttribute('data-track-click') 
+        && target.getAttribute('data-track-click').indexOf('true') > -1
+
+  // therem ight be an edge case where a manually tagretted link could also be an external link
+  // and we only wan't to fire one function in such a case
+  // hence the logic below
+  if (isExternalLink && isManuallyTargetedElement) {
+    this.trackTargettedLinkClick(target)
+  }
+  
+  if (!isExternalLink && isManuallyTargetedElement) {
+    this.trackTargettedLinkClick(target)
+  }
+
+  if (isExternalLink && !isManuallyTargetedElement) {
+    this.trackExternalLinkClick(target)
+  }
+}
+
+EventTracking.prototype.trackTargettedLinkClick = function (element) {
+  var eventParams = {},
+    category = element.getAttribute('data-track-category'),
+    action = element.getAttribute('data-track-action'),
+    label = element.getAttribute('data-track-label')
+
+  if (label) {
+    eventParams.event_label = label
+  }
+
+  if (category) {
+    eventParams.event_category = category
+  }
+  
+  this.sendEvent(action, eventParams)
+}
+
+EventTracking.prototype.trackExternalLinkClick = function (element) {
+
+  var eventParams = {},
+      action = element.textContent
+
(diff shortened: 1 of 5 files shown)


src/frontend/javascript/event-tracking.js
```diff
@@ -0,0 +1,96 @@
+function EventTracking () {
+  this.$bodyElement = document.querySelector('body')
+  this.$mainHeading = this.$bodyElement.querySelector('h1')
+  this.errorStatusCode,
+  this.isErrorPage = this.$mainHeading ?
```
AP-Hunt, line 7: We should extract this to a method. It's a sufficiently complicated expression.

src/frontend/javascript/event-tracking.js
```diff
@@ -0,0 +1,96 @@
...
+EventTracking.prototype.init = function() {
+
+  this.$bodyElement.addEventListener('click', this.handleClickEvent.bind(this))
+
+  if (this.isErrorPage) {
+    this.errorPageEvent(this.errorStatusCode)
+  }
+}
+
+EventTracking.prototype.handleClickEvent = function (e) {
```
AP-Hunt, line 23: Since we're in a Typescript project, we should make use of the type annotations

AP-Hunt (changes requested):

src/frontend/javascript/event-tracking.js
kr8n3r: this is client-side JS so we're not shipping TS or compiling from es6 to es5

src/frontend/javascript/event-tracking.js
kr8n3r: @AP-Hunt  extracted into a separate function

src/frontend/javascript/event-tracking.js
AP-Hunt: Aaah! Ok, gotcha.

src/frontend/javascript/event-tracking.js
```diff
@@ -0,0 +1,117 @@
...
+EventTracking.prototype.sendEvent = function (action, options) {
+  window.dataLayer = window.dataLayer || []
+  var gtag = function () {
+    dataLayer.push(arguments)
+  }
+
+  gtag('event', action, options)
+}
+
+EventTracking.prototype.checkForErrorPage = function(indetifier) {
```
AP-Hunt, line 92: s/indentifier/identifier/

src/frontend/javascript/event-tracking.js
```diff
@@ -0,0 +1,117 @@
...
+      statusCode: '500'
+    }
+  ]
+  if (!indetifier) return
+  
+  var errorPage  = errorPagesArray.filter(function(o){
+    return o.title === indetifier.textContent; 
+  });
+        
+  return errorPage? errorPage[0] : null;
```
AP-Hunt, line 114: Should this method be returning a bool? The use of it suggests it should, be the name suggests otherwise?

src/frontend/javascript/event-tracking.js
```diff
@@ -0,0 +1,117 @@
...
+      statusCode: '404'
+    },
+    {
+      title: 'Sorry an error occurred',
+      statusCode: '500'
+    }
+  ]
+  if (!indetifier) return
+  
+  var errorPage  = errorPagesArray.filter(function(o){
```
AP-Hunt, line 110: If it needs to return a bool, [the `some` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) would be better.

src/frontend/javascript/event-tracking.js
kr8n3r: it should return object data if condition met as we use that data later. Will find a better name

src/frontend/javascript/event-tracking.js
kr8n3r: will address the name as it return data if condition met

src/frontend/javascript/event-tracking.js
kr8n3r: update the logic for checking error page and firing events

src/frontend/javascript/event-tracking.js
kr8n3r: update the logic for checking error page and firing events

src/frontend/javascript/event-tracking.js
kr8n3r: spelling :(

AP-Hunt (approved):

PR #1445, opened 2021-05-11
Fix npm start scripts for cross-platform

0atman:
What
----

`npm start` used `source` lines, which are system-shell specific and break on (at least) ubuntu. 

How to review
-------------

make sure 
 - npm run start:stub-api
 - npm run start:with-stub-api

Both still work on your machine.

Who can review
---------------

JS devs

---


kr8n3r (approved):
runs beautifully on unix

PR #1456, opened 2021-05-13
Retire the  print stylesheet

kr8n3r:
What
----
Remove the print stylesheet from import and dditionally also change 
stylesheet media from `screen` to `all`, so we can:
- use govuk-frontend provided print styles
- additionally use our own  @print block in components

GOV.UK Front provides a utility class `govuk-!-display-none-print` to remove those elements from print.

Adding those classes to elements makes it clear that they are not meant to be visible in the print output.

If we later deem that we want to additionally prettify print output, we can remove those and move overrides into the @print block in the main stylesheet.

### Org page

**Before**

Screenshot 2021-05-13 at 17 22 47

**After**

Screenshot 2021-05-13 at 17 19 38

### Statement page

**Before**
Screenshot 2021-05-13 at 17 22 28

**After**

Screenshot 2021-05-13 at 17 20 29

How to review
-------------

- run locally and print selected page
- agree/disagree with the approach

Who can review
---------------
not @kr8n3r 
---


paroxp (approved):
Beautiful!

PR #1580, opened 2021-06-22
Move from `moment` to native `Date` object backed with `date-fns`

paroxp:
What
----

`moment` is EOL. We should be moving off it.

`moment`, although has served its purpose, has been somewhat opinionated and specific, which caused our codebase to purely rely on `moment.Moment` and `moment.Duration` types. This makes it difficult to move off, upgrade or maintain the code.

We've had a chat, and decided that native `Date`, should be preferably used throughout the codebase, to maintain the simplicity and potential for us to migrate/maintain when necessary.

I've started off, with writing few simple functions, such as `format`, `add`, `subtract`, `duration` - just to make the usage of `Date` a little simpler. But that list started to expand slowly and realistically... has been solved elsewhere. So instead of wasting time, I took a smaller steps and the time, and opted in for `date-fns`. Fell in love on the way.

As part of this, while manually having to change the usage of `moment`, my eyes were hurt by some linting highlights, so I've fixed them on the way.

How to review
-------------

- Sanity check
- Code review
- Have a look around a running instance (specifically around metrics please)


kr8n3r:
service metrics filter range on first view (last 24 hours):
 - end date should be today, but it's today - 1

Screenshot 2021-06-22 at 10 50 12

src/components/performance/views.tsx
```diff
@@ -81,7 +82,7 @@ export function MetricPage(props: IMetricPageProperties): ReactElement {
 
         <p className="govuk-body">
           Each point on the charts is aggregated data over {}
-          <strong className="non-breaking">{props.period.humanize()}</strong>.
+          <strong className="non-breaking">{formatDuration(props.period)}</strong>.
```
kr8n3r, line 85: we now say `1 week`, previously we said `7 days`. i'm fine with that

paroxp:
> end date should be today, but it's today - 1

I believe this is now fixed :)

> Each point on a graph is aggregated over 300 seconds of data, whereas on prod we convert that to 5 minutes (minor thing i reckon)

Actually that's a good spot. If this was to be 10000000s it would still not normalise... I've implemented a hack to take care of it.

> not sure if I can trust the graph start point (issue with stub data?)

I would hope so 😬 

> we now say 1 week, previously we said 7 days. i'm fine with that

I think I'm cool with that too 🤔

src/components/service-metrics/controllers.tsx
```diff
@@ -44,69 +44,98 @@ import {
...
+    obj.year || now.getFullYear(),
+    obj.month ? obj.month - 1 : now.getMonth(),
+    obj.day || now.getDate(),
+    obj.hour !== undefined ? obj.hour : now.getHours(),
+    obj.minute !== undefined ? obj.minute : now.getMinutes(),
+    obj.second !== undefined ? obj.second : now.getSeconds(),
+  );
+}
+
+export function parseRange(start: Range, stop: Range): IRange {
```
kr8n3r, line 109: - to note: there is now a minor change to populating range inputs if fields are left empty
previously we populated with current time (minus X) but now we populate with 00:00
that's fine, we still handle a broken user submission

- on load we populate time close to now, whereas before we used to do minus 1 hour
 which one is expected? see below

PROD
Screenshot 2021-06-22 at 11 53 56

DEV
Screenshot 2021-06-22 at 11 54 06

kr8n3r:
I'd like to test generated graphs and CSV download where data is more reliable. Can we deploy to somewhere?

src/components/service-metrics/controllers.tsx
paroxp: > on load we populate time close to now, whereas before we used to do minus 1 hour
which one is expected?

I don't know if there was ever an agreement about this if I'm honest... My thinking CURRENTLY is that the `end time` should be set to `new Date()`, where as the `start time` should be `sub(new Date(), { hours: 1 })`.

Do you have any preference?

kr8n3r: yep agreed, on load have `now` and `now-1h`

kr8n3r: when deployed it'a as per production

kr8n3r (approved):
massive piece of work 💯

PR #1586, opened 2021-06-23
Always format date given by prometheus

paroxp:
What
----

Prometheus is returning the date in various different formats. For
instance, once it comes back as ISO string, the other as a UNIX
timestamp. This is causing problems with `Date` object where as moment
has been magically figuring this out.

How to review
-------------

- Deploy to staging environment
- See if performance dashboard is now behaving

---


kr8n3r:
i suspect it's failing because of test data. we're already formatting the stub data to ISO in `getGappyRandomData` so I guess trying to format again in the controller isn't making it happy. wonder what format prom is returning in

kr8n3r (approved):
💯

PR #1659, opened 2021-07-12
Fix authentication cell markup on team members page

richardTowers:
What
----

Previously, users with "password" authentication got a visually hidden
"Authentication method:" span and a confusing link (to the same place as
the "Email address" cell one column to the left). Users with "Google" or
"Admin-google" don't get the link, and also don't get the visually
hidden span (which is probably an a11y issue).

Users where `props.privileged` was falsey got a table with an empty
column heading ("" instead of "Authentication") containing a bunch of
<NoTick />s (which are visually hidden, but readable by screen readers).
This changes the logic to skip the entire column (both the heading and
the cells) if the user isn't privileged enough to see the
Authentication.

How to review
-------------

Run locally, check the markup on the team members page is better.

Who can review
---------------

PaaSians

---

### DIFF
--- a/src/components/org-users/views.tsx
+++ b/src/components/org-users/views.tsx
@@ -714,9 +714,13 @@ export function OrganizationUsersPage(
                 <th className="govuk-table__header name" scope="col">
                   Email address
                 </th>
-                <th className="govuk-table__header authentication" scope="col">
-                  {props.privileged ? 'Authentication' : ''}
-                </th>
+                {props.privileged ? (
+                  <th className="govuk-table__header authentication" scope="col">
+                    Authentication
+                  </th>
+                ) : (
+                  <></>
+                )}
                 <th className="govuk-table__header is-org-manager" scope="col">
                   Org manager
                 </th>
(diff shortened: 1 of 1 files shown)


kr8n3r:
Hi @richardTowers  thanks for this. 
Is the problem you're trying to solve, to remove the "redundant" password link (as the email link already takes them to that page?

Screenshot 2021-07-12 at 18 20 41

what are we expecting `props.userOriginMapping[guid]` to return?

richardTowers:
Hi Jani - yep, that was the thing that I noticed that prompted me to raise this.

While working on it I spotted a couple of other issues:

1) If a user isn't privileged enough to see the authentication status of other users, they get an empty column full of `<NoTick />`s (not noticeable to sighted users, but would be confusing for screen reader users)
2) Cells with "Password" in get the visually hidden prefix "Authentication method: Password", whereas other Authentication methods just get "Google" (again, only noticeable for screen reader users, but confusing).

Hopefully the latest commit fixes both of those, and removes the link (which I think is redundant / confusing).

richardTowers:
> props.userOriginMapping[guid]

It will be `uaa` if the user has a password, `google` or `admin-google` if they use SSO.

kr8n3r (approved):
> > what are we expecting props.userOriginMapping[guid] to return?
> 
> It will be `uaa` if the user has a password, `google` or `admin-google` if they use SSO.


thanks. 
I've tested this and it seems to be all good
Will ask for a 2nd review in the morning 

Screenshot 2021-07-12 at 19 00 18
Screenshot 2021-07-12 at 19 01 34

paroxp (approved):

PR #1704, opened 2021-07-30
[#178944607] calculator: add aws-sqs-queue services

risicle:
What
----

Made calculator aware of SQS pricing.

One thing is that of course it insists on listing the two different variants of SQS (standard & fifo), though it has no effect on pricing. A little ugly, but how much do we care? Would probably take some hackery to omit plan selection for specific offerings.

Screenshot 2021-07-30 at 14 58 58

How to review
-------------

You'll probably first have to deploy `paas-cf/ris-sqs-pricing-178944646` (https://github.com/alphagov/paas-cf/pull/2713) to a dev paas, then hack the create-cloudfoundry pipeline to use this branch for deployment and do a deploy of that.

Alternatively look @ dev01 before the day is out (2021-07-30).

Who can review
---------------

Anyone technically capable to try it or irresponsible enough to just trust me.

---

### DIFF
--- a/src/components/calculator/controllers.test.tsx
+++ b/src/components/calculator/controllers.test.tsx
@@ -85,6 +85,11 @@ describe('calculator test suite', () => {
             name: 'aws-s3-bucket default',
             plan_guid: 'f4d4b95a-f55e-4593-8d54-3364c25798c9',
           },
+          {
+            ...defaultPricingPlan,
+            name: 'aws-sqs-queue standard',
+            plan_guid: 'f4d4b95a-f55e-4593-8d54-3364c25798c9',
+          },
         ]),
       );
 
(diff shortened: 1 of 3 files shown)


kr8n3r:
it's probably defined in paas-billing, but it would be good to say something like Queue type: <type> in the select

risicle:
All good questions.

My assumption was that it's not so easy to "remove the dropdown" without a bunch of special-case code because the calculations actually operate on _plans_ rather than _offerings_, and they're separate plans. Could have some sort of "list of plans to ignore", ignore one of the SQS plans so that the entry ends up behaving like the S3 line?

risicle:
See new commit.

Result: 
Screenshot 2021-07-30 at 16 51 12

kr8n3r (approved):
👏

kr8n3r:
needs https://github.com/alphagov/paas-cf/pull/2713

PR #1787, opened 2021-08-31
Upgrade typescript and fix unknown ypes

kr8n3r:
What
----

In Typescript 4.4.2 there is a breaking change https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#using-unknown-in-catch-variables

Our acceptance tests that run on staging and prod fail.

This adds type `any` to variable in the try/catch block.

more info:
- https://github.com/microsoft/TypeScript/pull/41013
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

How to review
-------------

- if you have a dev env, deploy there and run the acceptance test
-  alternatively, obtain envs from cf and credhub, export variable to point to dev01 or dev02 and run `npm run test:acceptance`

Who can review
---------------

not @kr8n3r 

---

### DIFF
--- a/src/lib/uaa/uaa.ts
+++ b/src/lib/uaa/uaa.ts
@@ -192,7 +192,7 @@ export default class UAAClient {
     try {
       const response = await this.request('get', `/Users/${userGUID}`);
       return response.data as types.IUaaUser;
-    } catch (err) {
+    } catch (err: any) {
       if (err.response.status === 404) {
         return null;
       }
--- a/acceptance-tests/main.test.ts
+++ b/acceptance-tests/main.test.ts
@@ -40,7 +40,7 @@ describe('paas-admin', () => {
       try {
         await axios.request({ maxRedirects: 0, url: PAAS_ADMIN_BASE_URL });
       }
-      catch(rejection) {
+      catch(rejection: any) {
         expect(rejection.response.status).toEqual(302);
       }
     });
--- a/package-lock.json
+++ b/package-lock.json
@@ -123,7 +123,7 @@
         "ts-jest": "^27.0.5",
         "ts-loader": "^9.2.5",
         "ts-node": "^10.2.1",
-        "typescript": "^4.3.5",
+        "typescript": "^4.4.2",
         "webpack": "^5.38.1",
         "webpack-cli": "^4.8.0",
         "webpack-node-externals": "^3.0.0"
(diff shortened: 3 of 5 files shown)


src/lib/uaa/uaa.ts
```diff
@@ -192,7 +192,7 @@ export default class UAAClient {
     try {
       const response = await this.request('get', `/Users/${userGUID}`);
       return response.data as types.IUaaUser;
-    } catch (err) {
+    } catch (err: any) {
```
paroxp, line 195: Instead of `any`, do you think this could be `AxiosError`?

https://github.com/axios/axios/blob/master/index.d.ts#L92-L99

kr8n3r: @paroxp that would make sense, should have checked Axios'types
though, it's not happy 
Screenshot 2021-08-31 at 12 23 01

kr8n3r: ah more info at https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

paroxp: Weird. As you were! Thanks for checking this out.

kr8n3r: Looking at example they have
 
```
if (err instanceof Error) {
     console.error(err.message);
 }
```
which wouldn't exactly work in our case, but i we can somehow defien a custom type that extends `AxiosError` ?

kr8n3r:
@paroxp what do you think of changes in https://github.com/alphagov/paas-admin/pull/1787/commits/48079c91c888853183b2f0958d6c31a28ec13634 ?

as we can't do `catch ( err: AxiosError)` this is the best i could come up to check that type is `AxiosError`

Screenshot 2021-09-01 at 14 09 53

paroxp:
Yeah, this looks good. I imagine you still need `any` or `unknown` type. And if I think about it, there could be number of errors that could be thrown. Us doing the checks and resolving to default behaviour if it's not something we expected is OK I think.

kr8n3r:
> Yeah, this looks good. I imagine you still need `any` or `unknown` type. And if I think about it, there could be number of errors that could be thrown. Us doing the checks and resolving to default behaviour if it's not something we expected is OK I think.

`unknown` is now the default with typescript 4.4.2 and `--strict`

as tests are now failing will have to look at a different solution, might have to be something so I can do 
`err instance of <customErrorThatExtendsAxiosError>`

kr8n3r:
@paroxp going back to previous `any` type for the error as it's become to cumbersome

PR #1833, opened 2021-09-16
Add last logon infromation to organisation user list

kr8n3r:
What
----

For privileged users (admin, org manager) we show information when was the last time the user logged in. This information comes from [UAA](https://docs.cloudfoundry.org/api/uaa/version/75.7.0/index.html#get-3)

How to review
-------------

- run locally or deploy to a dev env
- test pass
- check code makes sense

Visual changes
---------------

**Privileged user view**
Screenshot 2021-09-16 at 16 55 10

**Regular user view**

Screenshot 2021-09-16 at 15 59 41

Who can review
---------------

not @kr8n3r 

---

### DIFF
--- a/src/components/org-users/views.tsx
+++ b/src/components/org-users/views.tsx
@@ -1,4 +1,6 @@
 import React, { ReactElement, ReactNode } from 'react';
+import { format } from 'date-fns';
+import { DATE_TIME } from '../../layouts';
 
 import { capitalize } from '../../layouts';
 import { NoTick, Tick } from '../../layouts/partials';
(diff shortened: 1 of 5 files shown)


src/components/org-users/views.tsx
```diff
@@ -803,6 +819,17 @@ export function OrganizationUsersPage(
                       ))}
                     </ul>
                   </td>
+                  {props.privileged ? (
+                    <td className="govuk-table__cell">
+                      {props.userExtraInfo[guid].lastLogonTime ? (
+                        format(new Date(props.userExtraInfo[guid].lastLogonTime!), DATE_TIME)
```
0atman, line 825: @kr8n3r This line is a bit dense, doesn't `format()` take an int? and does adding `!` cast to bool?

kr8n3r: you are correct. seem likr i forgot to update when i was testing outputs from the controller. have updated
as for ! - to stop it complaining about 

```
Argument of type 'number | undefined' is not assignable to parameter of type 'number | Date'.
  Type 'undefined' is not assignable to type 'number | Date'
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

0atman: Aha! I don't feel so bad about not knowing a typescript operator :grin: Thanks for the link!

kr8n3r: don't beat yourself up to hard :) you were probably think of `!!`

0atman (approved):
Can't get the tests to run, but changes look good. :ok_hand: 
Screenshot from 2021-09-22 12-55-52

PR #2070, opened 2021-11-19
[#179507076] Make Zendesk tickets convey urgency when a service is down

AP-Hunt:
What
----

In the past we've missed service down events from tenants because the ticket
subject was identical to less severe problems. By changing the subject when the
severity is high is enough, we hope to avoid that in future.

How to review
-------------
1. Code review
2. In the test I've written, is that the correct way to test the content of a POST request to another service?

---

### DIFF
--- a/src/components/support/controllers.tsx
+++ b/src/components/support/controllers.tsx
@@ -431,6 +431,12 @@ export async function HandleSomethingWrongWithServiceFormPost(
 
   const client = zendesk.createClient(ctx.app.zendeskConfig);
 
+  let subject = ""
+  if(body.impact_severity === "service-down") {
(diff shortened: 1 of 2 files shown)


src/components/support/controllers.tsx
```diff
@@ -431,6 +431,12 @@ export async function HandleSomethingWrongWithServiceFormPost(
 
   const client = zendesk.createClient(ctx.app.zendeskConfig);
 
+  let subject = ""
+  if(body.impact_severity == "service-down") {
```
kr8n3r, line 435: ===

tomrosier (approved):

kr8n3r (approved):

PR #2125, opened 2021-12-08
Filter out more unreal orgs

kr8n3r:
What
----

It seems we have some test orgs that begin with CAT not CATS that weren't filtered out.
stray test org

This update to the regex and extended test resolve it.

How to review
-------------

- check code
- tests pass

Who can review
---------------

not @kr8n3r 

---


tomrosier (approved):

AP-Hunt (approved):
Nice catch

PR #2134, opened 2021-12-10
Show expiration notice / days count for trial organisations

kr8n3r:
What
----

For trial organisations show text depending on number of days left of the trial period as follows:

- between 1 and 90 days (trial period duration): "Trial period expires in X day(s)"
- 0 and negative days (more than 90 days have passed): "Trial period has expired."

If an org is a trial org (it's quota is still "default") we calculate the number of calendar days between current date and when organisation was created.

If an org is a billable org, then we pass null to the view and no text is displayed.

Visual changes
-------------
Screenshot 2021-12-10 at 11 15 53
Screenshot 2021-12-10 at 11 27 31

How to review
-------------

- run locally and manipulate test data
- or deploy to dev 01/02

Who can review
---------------

not @kr8n3r

---

### DIFF
--- a/src/components/spaces/controllers.tsx
+++ b/src/components/spaces/controllers.tsx
@@ -1,6 +1,7 @@
 import lodash from 'lodash';
 import React from 'react';
 import { validate as uuidValidate } from 'uuid';
+import differenceInCalendarDays from 'date-fns/differenceInCalendarDays'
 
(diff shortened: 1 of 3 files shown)


cadmiumcat (approved):
Nice 🙌 
Code makes sense. Tried on dev env and looks good

PR #2291, opened 2022-01-30
Replace method of generating UUIDs

kr8n3r:
What
----

Replace the use of `uuid` package for generating and validating UUIDs with NodeJS `crypto.randomUUID` and custom test function to check UUID validity.

Why
----

We'd like to reduce the number of dependencies we use for maintenance reasons and take advantage of native methods (if available) or own code if feasible. 

Added bonus: `randomUUID` is marginally faster, but negligible for our use-case

Assumption 
----

- we only need to use [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) (which is what `crypto.randomUUID` generates)

How to review
-------------

- review by commit
- run locally or deploy to dev env

Who can review
---------------

not @kr8n3r 

---

### DIFF
--- a/src/layouts/helpers.tsx
+++ b/src/layouts/helpers.tsx
@@ -72,3 +72,10 @@ export function conditionallyDisplay(
 export function capitalize(value: string): string {
   return value.charAt(0).toUpperCase() + value.slice(1);
 }
+
+/* Check if string is a valid UUID */
+export function checkIfValidUuid(input: string): boolean {
+  // 8-4-4-4-12 arrangement of 36 characters
+  const regexExp = /^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i;
+  return regexExp.test(input.toLowerCase());
+}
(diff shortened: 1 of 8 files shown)


src/layouts/helpers.tsx
```diff
@@ -72,3 +72,11 @@ export function conditionallyDisplay(
 export function capitalize(value: string): string {
   return value.charAt(0).toUpperCase() + value.slice(1);
 }
+
+/* Check if string is a valid UUID */
+export function checkIfValidUuid(input: string):boolean {
+  // 8-4-4-4-12 arrangement of 36 characters
+  const regexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi;
```
AP-Hunt, line 79: If you cast the string to upper case, you could simplify the regular expression

paroxp (approved):
Yes.

Let's kill all the dependencies.

src/layouts/helpers.tsx
paroxp: Nice touch Andy. Better for readability of "write-only" regex's.

kr8n3r: this in mind?

AP-Hunt: Precisely what I was thinking, yeah

PR #2545, opened 2022-04-08
Test esm govuk frontend

kr8n3r:
**below error is solved by  adding .mjs extension to all component imports and .js extension to all polyfill imports in the govuk-esm folder**

with https://github.com/alphagov/paas-admin/pull/2545/commits/113bd54e8d72ffdeb5bc1493b9425512a6dbdf35

the errors from govuk-frontend (for all imports)

```
ERROR in ./node_modules/govuk-frontend/govuk-esm/all.mjs 12:0-41
Module not found: Error: Can't resolve './components/tabs/tabs' in '/Users/janikraner/paas/paas-admin/node_modules/govuk-frontend/govuk-esm'
Did you mean 'tabs.mjs'?
BREAKING CHANGE: The request './components/tabs/tabs' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./src/frontend/javascript/init.js 1:0-39 4:6-12
```


config/webpack.config.js
```diff
@@ -52,7 +52,8 @@ let cfg = {
   },
 
   resolve: {
-    extensions: ['.ts', '.tsx', '.js', '.json'],
+    fullySpecified: false,
```
vanitabarrett, line 55: @kr8n3r I've had some time to dig into this and I think `fullySpecified` needs to be defined in a resolve block within the `module: { rules: [] }` section below, like so:

```
module: {
    rules: [
      {
        test: /\.mjs$/,
        type: 'javascript/auto',
        resolve: {
          fullySpecified: false
        }
      }
    ]
}

If you do this, I think you can remove the hardcoded file extensions in the govuk-esm javascript files.

https://webpack.js.org/configuration/module/#resolvefullyspecified

PR #2564, opened 2022-04-14
Modify org users table display

kr8n3r:
What
----

[There are cases](https://admin.london.cloud.service.gov.uk/organisations/d96a0801-7a7c-4706-b546-cf4ca9b4ad31/users) where an org might not contain any members.
Currently we still show the table header which is not right.

This PR checks if any users exists and if not, does not render the table at all.

### Before

Screenshot 2022-04-14 at 09 41 20

### After

Screenshot 2022-04-14 at 09 41 03

How to review
-------------

- check test passes
- can deploy locally or to a dev env and remove all users from an org to check

Who can review
---------------
not @kr8n3r 

---


AP-Hunt (approved):
I like this change a lot

PR #2808, opened 2022-06-24
Add requester details to zendesk support ticket if logged in

kr8n3r:
What
----

If the requester of our support forms is logged into paas-admin, we want to include their cloudfoundry details in the ticket as a private comment, so that we can quickly check that they have appropriate roles.

If they're not logged in or don't have an account, we also include that in the comment.

On form submission we check the session for user token and check UAA if the user exists. If not we, return an empty object.

If they do, we do a lookup if they're members of any organisations and if so what are their roles.

Once all the information is collated, we first create the zendesk ticket and then update it with the requester details in the private comment.

How to review
-------------

- best reviewed by commits
- can deploy to a dev env to test scenarios (or look at screenshot below)

Who can review
---------------

not @kr8n3r 

---
Not logged in
Screenshot 2022-07-04 at 10 01 46

Logged in but no roles
Screenshot 2022-07-04 at 10 03 07

All the data
Screenshot 2022-07-04 at 10 05 13

### DIFF
--- a/src/components/support/controllers.tsx
+++ b/src/components/support/controllers.tsx
@@ -6,6 +6,9 @@ import { IParameters, IResponse } from '../../lib/router';
 import { IContext } from '../app';
 import { IDualValidationError, IValidationError } from '../errors/types';
 
+import UAAClient from '../../lib/uaa';
+import { Token } from '../auth';
+
 import {
   ContactUsPage,
   DocumentsCrownMoU,
@@ -27,6 +30,12 @@ import {
   supportFormFieldsText,
   SupportSelectionPage,
 } from './views';
+import CloudFoundryClient from '../../lib/cf';
+
+interface IUserTypedRequester {
+  readonly email: string;
+  readonly name: string;
+}
 
 interface ISupportFormName {
   readonly name?: string;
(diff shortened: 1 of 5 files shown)


src/components/support/controllers.tsx
```diff
@@ -71,10 +80,114 @@ interface ISignupForm extends ISignupFormValues {
...
+  const user = await uaa.getUser(token.userID);
+  // get their roles
+  const roles = await cf.userRoles(user!.id);
+  // fetch org details and roles the user has 
+  const userRoleTypeAndOrg = await Promise.all(
+    roles
+    .filter(r => r.relationships.organization.data)
+    .map(async r => ({
+      orgGuid: ((await cf.organization(r.relationships.organization.data.guid!)).metadata.guid),
+      orgName: ((await cf.organization(r.relationships.organization.data.guid!)).entity.name),
```
risicle, line 128: I take it the cf client doesn't do any sort of caching. Could we re-use this lookup and so save us half the API requests?

I assume we're looking up the org's guid ... via its guid ... so we can be sure it exists? Otherwise we'd surely omit the request.

kr8n3r: cache no, but you're correct. we can just used the guid in the `data`  object.
amended

PR #3031, opened 2022-08-30
Update org quota display

kr8n3r:
What
----

- Add name of assigned quota
- Update tests

If viewer has `org manager` role: show detailed breakdown as before, else only show the assigned quota name.

Why
----

Information display differs depending on viewing user's role. Only for org managers are resources calculate correctly.

Visual changes
---------------

### Before
**org manager view**

Screenshot 2022-08-31 at 08 13 13

**regular view**

Screenshot 2022-08-31 at 08 12 49

### After

**org manager view**

Screenshot 2022-08-31 at 08 15 56

**regular view**

Screenshot 2022-08-31 at 08 15 31

How to review
-------------

- tests pass
- review code
- optionally deploy to an env, and view with different roles

Who can review
---------------

not @kr8n3r 

---

### DIFF
--- a/src/components/spaces/views.test.tsx
+++ b/src/components/spaces/views.test.tsx
@@ -351,10 +351,6 @@ describe(SpacesPage, () => {
     );
     expect(container.querySelectorAll('table tbody tr')).toHaveLength(2);
     expect(container.querySelector('h1')).toHaveTextContent(organization.entity.name);
-    expect(container.querySelector('h2 + span')).toHaveTextContent('20.0%');
-    expect(container.querySelector('p')).toHaveTextContent(
-      'Using 1.00 GiB of memory out of a maximum of 5.00 GiB.',
-    );
     expect(queryByText('View team members')).toBeTruthy();
     expect(queryByText('Trial')).toBeTruthy();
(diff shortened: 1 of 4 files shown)


src/components/spaces/views.test.tsx
```diff
@@ -382,6 +378,44 @@ describe(SpacesPage, () => {
...
+        isBillingManager={false}
+        organization={organization}
+        spaces={[space, { ...space, quota: undefined }]}
+        users={[null]}
+      />,
+    );
+    expect(container.querySelectorAll('.org-summary-stat__value')[0]).toHaveTextContent('default');
+    expect(container.querySelectorAll('.org-summary-stat__value')[0]).not.toHaveTextContent('20.0%');
+    expect(container.querySelector('p')).not.toHaveTextContent(
+      'Using 1.00 GiB of memory out of a maximum of 5.00 GiB.',
```
risicle, line 415: Sometimes when negating assertions it's useful to be _less_ exact/specific with the condition. This for example would "pass" if the string were present but just missing the final full stop.

kr8n3r: good point. I've update the test to check for presence of that container

risicle (approved):
WFM on dev03

PR #3121, opened 2022-09-23
Fix password reset error handling

kr8n3r:
What
----

Wrap the password reset request in a `try...catch` block so that when there is an error in the response, we show the error in the form instead of the the "Sorry there's been a problem" page.

As discovered in https://govuk.zendesk.com/agent/tickets/5079686, UAA does not like if users try to reset the password that matches the existing password, returning
```
UAAClient: post /password_change failed with status 422 and data {"error_description":"Your new password cannot be the same as the old password.","error":"invalid_password","message":"Your new password cannot be the same as the old password."}
```

We did not cater for this scenario. This change

- exposes the error object in the UAA response if it exists so that, we can use it in the form
- wraps the password reset request in a `try...catch` block so that when there is an error in the response, we show the error in the form with the message from the response

How to review
-------------

- try to reset user password with the same password
- deploy to dev env and compare with current prod

Who can review
---------------

not @kr8n3r 

New error message
----
Screenshot 2022-09-23 at 11 34 05

---

### DIFF
--- a/src/components/users/controllers.test.tsx
+++ b/src/components/users/controllers.test.tsx
@@ -532,6 +532,30 @@ describe(users.resetPassword, () => {
     expect(response.status).toEqual(400);
     expect(response.body).toContain('should contain a special character');
   });
+
+  it('should throw an error if the new password is the same as old password', async () => {
+
+    nockUAA
+      .post('/oauth/token?grant_type=client_credentials')
+      .reply(200, '{"access_token": "FAKE_ACCESS_TOKEN"}')
+
+      .post('/password_change')
+      .reply(422, {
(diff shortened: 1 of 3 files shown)


paroxp (approved):
`password-STR0NG-like-j3nk1n$` ... this will haunt us forever.

Looks good, It's a shame we have to wrap it in the try-catch. But I cannot remember what we've done before (I guess a middleware catching these?).

Looks good either way.

kr8n3r:
it's currently caught by middleware in routes which just brings up the generic error page

paroxp:
Ah. Yeah, that's lame. Improvement it is!

PR #3377, opened 2022-11-25
Replace pegjs and loader with vendored pegjs parser

kr8n3r:
What
----

`Pegjs` and especially `pegjs-loader` aren't well maintained leading to https://github.com/alphagov/paas-admin/security/dependabot/27 which we cannot force an upgrade of.

Instead, we're vendoring the pegjs parser output generated by https://pegjs.org/online from our grammar rules.

Now grammar rules aren't imported and transformed by webpack; instead the already transformed rules are used.

This is a tradeoff between security and ease of maitenance, but since we haven't changed the grammar since 2019 there should be no issues. It's trivial to just recompile the exiting grammar rules using pegjs online compiler.

How to review
-------------

- tests pass
- /calculator page maths still works

Who can review
---------------

not @kr8n3r 

---


AP-Hunt (approved):
I think this is a very sensible approach.

PR #4575, opened 2023-10-30
Don't copy objects, with large files it is too slow

fearoffish:
What
----

Large events files returned by the billing app can take 100s of seconds to reduce, due to the amount of object copying taking place. This change overwrites the objects with new values instead. Not the best programming choice, but in my tests with a sample large json file it reduced the test run from 2000 seconds to 41 seconds.

How to review
-------------

👀  on integration tests

Who can review
---------------

Anyone with admin access.

---


risicle:
If you're now mutating `resources` in the lambda, I'd avoid calling it as a `reduce` because that's no longer really what it does. You can surely just run a normal foreach/map/whatever js calls it in 2023 to achieve this.

Issue #2483, opened 2022-03-23
Figure of what to do with Jest and ES Modules


kr8n3r:
## What

**TLDR; We need to find a way to make ES module dependencies work with Jest and TS-jest without affecting our javascript bundle meant for the client-side browser.**

We have an increasing [number of dependencies](https://github.com/alphagov/paas-admin/labels/type%3A%20module) moving to [ES Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)
They link to [Sindre Sorhus's write-up](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)
 
 Tests now break in Jest test with 

```
Jest encountered an unexpected token

 Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

By default "node_modules" folder is ignored by transformers.

 Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

You'll find more details and examples of these config options in the docs:
 - https://jestjs.io/docs/configuration
 - For information about custom transformations, see:
 - https://jestjs.io/docs/code-transformation
```

For use with Typescript use use a jest wrapper, `ts-jest`and they have [docs on ES module support](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/)

In most cases we already use `import` instead of `require`.

## Why 

So that we can update dependencies that have move to ES modules

## It’s done when 

* tests pass and update pull requests are merged

## If we don’t do this 

We'll be stuck with out of date dependencies.

Issue #1666, opened 2021-07-14
Jest leaks memory catastrophically  


0atman:
![jest](https://user-images.githubusercontent.com/114097/125645592-c7e766bf-8b5c-4f87-8d69-0aaac1f56240.png)

my memory useage goes from 4GB up to my computer's limit of 16GB and then eats all the swap space (2GB) and then my machine freezes up
![Screenshot from 2021-07-14 13-51-26](https://user-images.githubusercontent.com/114097/125645609-132f98a4-44c2-4de8-8925-06e8e3db90f5.png)

More reading:
- https://chanind.github.io/javascript/2019/10/12/jest-tests-memory-leak.html
- https://github.com/kulshekhar/ts-jest/issues/1967

kr8n3r:
we no longer use Jest

PR #4956, opened 2024-02-27
Replace Jest & Nock with Vitest & MSW and move to ES module output

kr8n3r:
What
----

Replace Jest & Nock with [Vitest](https://vitest.dev/) & [Mock Service Worker](https://mswjs.io/) and move to an ES module build
Update dependencies

Closes: https://github.com/alphagov/paas-admin/issues/2483 https://github.com/alphagov/paas-admin/issues/1666

Why
----

Current setup makes it almost impossible to update dependencies that have migrated from CommonJS to ES Modules as Jest failed to work properly.

Replacing Jest with Vitest (which uses vite as a builder (and is ES module native) solved that. Nock is a CommonJS library so it wasn't working Vitest so a switch to MSW was needed.

Finally as Vite will be moving away from CJS build soon, we needed to also build our app an an ES module.
In doing so, we have also been able to update a bunch of dependencies

How to review
-------------

- tests pass and compare [new tests](https://github.com/alphagov/paas-admin/actions/runs/8298826176/job/22713100583#step:6:232) with [tests on main](https://github.com/alphagov/paas-admin/actions/runs/8232599322/job/22510377187#step:6:1528) [781 vs 782 as 2 test were combined into 1)
- deploy to a dev env to check app is working

Who can review
---------------

not @kr8n3r 

---


dark5un:
I tried it on dev and works fine... however maybe these conflicting files need to be resolved.

dark5un (approved):
LGTM

kr8n3r:
after rebasing from main GHA tests keep failing. investigating

